PR #8, opened 2016-11-23
fill_keywords script to handle svn:keywords in git (SOFTWARE-2538)

edquist:
In the svn sources, there are several instances where the svn:keywords
property is used on files to replace things like $Revision$ with some
text, which actually gets parsed in some places.

The application of these keywords is not consistent across the svn
sources, as (apparently) historically files were copied around with
already-filled-in keywords rather than svn-copy'd, etc.

In any case, because some places in the probes actually parse these
keyword values and break if they are not filled in, we need to fill them
in somehow for the releases based on the git sources.

Perhaps someday we'll abolish the use of these keywords altogether  :)

### DIFF
--- a/build/fill_keywords
+++ b/build/fill_keywords
@@ -1,0 +1,55 @@
+#!/usr/bin/python
+
+# for the time being, we need to populate svn:keyword fields like $Revision$
+# in the git sources when packaging a release.
+
+import os
+from subprocess import Popen, PIPE
+
+def runcmd(cmdline):
+    return Popen(cmdline, stdout=PIPE).communicate()[0].rstrip("\n")
+
+def HeadURL(path):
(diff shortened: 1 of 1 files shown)


osg-cat:
I do not like adding code to work around a problem caused by archaic coding conventions. Let’s pursue the idea from your last sentence — check with Tanya and see if there is any reason why we cannot just strip out all of the code related to the keywords, parsing the keywords, etc.

bbockelm:
I'm with Tim on this one - can we just pick off the problem use cases?

edquist:
Ok, that sounds good.

For reference here, it looks like the only keyword that's actually used (other than to decorate files) is `$Revision$`, which is parsed and sent to Gratia, and ends up in the Gratia db in `ProbeDetails_Xml.rawxml` in the `Reporter` and `ReporterLibrary` xml elements.

PR #9, opened 2016-12-15
Set the host description to the ProbeConfig SiteName for local jobs

brianhlin:
In the case where a VO needs to account for locally run
jobs (i.e. SuppressGridLocalRecords="0"), we need to set the host
description to an OIM-registered name. In our documentation we tell
our users to set the SiteName in their ProbeConfig to the Resource
Name for their submit host so we use that.

### DIFF
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -551,10 +551,13 @@ def determine_host_description(classad):
 
     If there's no special host description, this returns None.
     """
-    if 'MATCH_EXP_JOBGLIDEIN_ResourceName' not in classad:
+    try:
+        if classad['MATCH_EXP_JOBGLIDEIN_ResourceName'] == 'Local Job':
+            host_descr = GratiaCore.Config.get_SiteName()
+        else:
+            host_descr = classad['MATCH_EXP_JOBGLIDEIN_ResourceName']
+    except KeyError:
         return None
-
-    host_descr = classad['MATCH_EXP_JOBGLIDEIN_ResourceName']
 
     # Check first for SE-based matching.
     if ('DESIRED_SEs' not in classad) or ('MATCH_GLIDEIN_SEs' not in classad):
--- a/condor/99_gratia-gwms.conf
+++ b/condor/99_gratia-gwms.conf
@@ -1,18 +1,23 @@
 #
-# This file specifies the classads that need to be included in each job for
-# proper accounting of GlideinWMS jobs.  If you experience any problems with
-# JOBGLIDEIN_ResourceName, or the gratia probe, please email
-# osg-software@opensciencegrid.org
+# This file specifies the classads that need to be included in each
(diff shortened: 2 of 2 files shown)


brianhlin:
@edquist could you take a look at this?

condor/condor_meter
```diff
@@ -551,11 +551,14 @@ def determine_host_description(classad):
 
     If there's no special host description, this returns None.
     """
-    if 'MATCH_EXP_JOBGLIDEIN_ResourceName' not in classad:
+    try:
+        if classad['MATCH_EXP_JOBGLIDEIN_ResourceName'] == 'Local Job':
+            host_descr = GratiaCore.Config.get_SiteName()
+        else:
+            host_descr = classad['MATCH_EXP_JOBGLIDEIN_ResourceName']
+    except KeyError:
```
edquist: @brianhlin, how would you feel about just using an if/elif/else here?  (something like [here](https://github.com/opensciencegrid/gratia-probe/blob/master/condor/condor_meter#L745))

I'm not really wild about the try/except thing (does a `KeyError` only come from `classad[]` or can `get_SiteName()` generate one too?) ... But if you're really feeling the try/except thing that works too.

condor/condor_meter
```diff
@@ -551,11 +551,14 @@ def determine_host_description(classad):
 
     If there's no special host description, this returns None.
     """
-    if 'MATCH_EXP_JOBGLIDEIN_ResourceName' not in classad:
+    try:
+        if classad['MATCH_EXP_JOBGLIDEIN_ResourceName'] == 'Local Job':
+            host_descr = GratiaCore.Config.get_SiteName()
```
djw8605: How will this work with SuppressGridLocal jobs?  Like dagman jobs?

condor/condor_meter
brianhlin: I'm not really attached to any particular way of doing it but I think it's safe in this case:

* If `get_SiteName()` throws a `KeyError`, is it worthy of stopping the entire probe run or should we just continue constructing the record as best we can? We could try setting the host description to `MATCH_EXP_JOBGLIDEIN_ResourceName`, which would be `Local Job` in that case, but I'm not sure if that's any better than setting it to `None` (and possibly worse because of how the `None` case is [handled](https://github.com/opensciencegrid/gratia-probe/blob/master/condor/condor_meter#L704-L708)).
* If `classad[]` throws a `KeyError`, we have no hope of knowing where the job ran so we should return `None`. We could feasibly set it to the ProbeConfig's SiteName but it wouldn't necessarily reflect reality.

condor/condor_meter
brianhlin: To be clearer: this change makes sure local job records have the right data. `SuppressGridLocalJobs` decides whether or not to upload the local job records.

edquist: @djw8605, are you satisfied here or do you think there will be issues still?

PR #10, opened 2016-12-19
Don't use svn:keywords anymore (SOFTWARE-2538)

edquist:
This removes remaining uses of `$Revision$` and `$Id$`-style svn:keywords that were being used in the svn version of gratia.

The only relevant bit is that the version of the probes does in fact get sent to the gratia collector.  Rather than try to reproduce the old behavior by filling these keyword fields in with revision information at build time, we will send the rpm VR string, which is actually more precise even than a git hash, since an rpm build may contain patches on top of the upstream sources.

I talked with @bbockelm and @djw8605 about this last week, who seemed to be happy with this plan.  But please comment if you think there are any issues.

### DIFF
--- a/boinc/boincprobe.py
+++ b/boinc/boincprobe.py
@@ -26,8 +26,7 @@ localjobid = ''
 endtime = ''
 user = 'Einstein@Home'
 
-rev = '$Revision: 3273 $'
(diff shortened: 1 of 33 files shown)


edquist:
I did a local build with these changes merged into `master` and did a full osg-test run of osg-tested-internal on an el6 VM with no failures.

PR #16, opened 2017-03-07
Fix eval in debug line (SOFTWARE-2636)

djw8605:


edquist:
@djw8605, I'm not really familiar with `classad.eval` vs `classad.__getitem__` ... it works right?  :)

djw8605:
Not sure.  I think classad can either be a htcondor python wrappers classad object, or the caselessDict defined above.  In this case, it was caselessDict, which does not have an `eval`.

edquist:
Ok, so in the case that it's a classad object, where `.eval` _is_ defined, what happens in the above line for `classad['CompletionDate']`? (is it the same as `classad.eval('CompletionDate')`?)

djw8605:
Ah, I see.  You are trying to figure out if we even need eval?  I'm not sure.  I assume this was done in case `CompletionDate` is an expression.  But I don't know when that would be the case.

edquist:
Also I'm just asking to confirm that `classad['CompletionDate'] > 0` won't blow up if it's a classad object.

But yeah if that expression is safe, then presumably `classad['CompletionDate']` evaluates to something numeric, and presumably `classad.eval('CompletionDate')` ... evaluates to the same thing?  Because if `classad['CompletionDate']` is an expression then I'm wondering if `classad['CompletionDate'] > 0` will even work.

If you think it's ok though, I'll take your word for it  :)

djw8605:
I would be more on the cautious side.  It didn't blow up before in tests and production, and only blew up now with `eval`.

edquist:
Sounds good - the current changes in any case shouldn't break anything new.

PR #17, opened 2017-06-05
Added the ability to submit to the register arbitrary class Ads

efajardo:
@bbockelm or @edquist could you review this, please? This was done for [SOFTWARE-2714](https://jira.opensciencegrid.org/browse/SOFTWARE-2714)

### DIFF
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -714,12 +714,12 @@ def classadToJUR(classad):
     ########################################################################################
     # Code added to send to arbitrary Ads SOFTWARE-2714
     ArbitraryJobAttrs = GratiaCore.Config.getConfigAttribute("ExtraAttributes")
-    DebugPrint(5, "ArbitraryJobAd: %s" % str(ArbitraryJobAttrs))
+    DebugPrint(5, "Arbitrary Job Attributes: %s" % str(ArbitraryJobAttrs))
     ArbitraryAttrslist = re.split(",?\s?", ArbitraryJobAttrs)
     #ArbitraryAttrslist = ArbitraryJobAds.split(",")
     for arbitraryAttr in  ArbitraryAttrslist:
         if arbitraryAttr in classad:
-            DebugPrint(5, "Arbitrary ad: %s found in classAd, value %s" % (arbitraryAttr, str(classad[str(arbitraryAttr)]))) 
+            DebugPrint(5, "Arbitrary attribute: %s found with value %s" % (arbitraryAttr, str(classad[str(arbitraryAttr)]))) 
             r.AdditionalInfo(arbitraryAttr, str(classad[str(arbitraryAttr)]))
     ########################################################################################
     if 'MachineAttrCpus0' in classad:
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -713,15 +713,14 @@ def classadToJUR(classad):
     ########################################################################################
     ########################################################################################
     # Code added to send to arbitrary Ads SOFTWARE-2714
-    ArbitraryJobAds = GratiaCore.Config.getConfigAttribute("ArbitraryJobAdAttributes")
-    DebugPrint(5, "ArbitraryJobAd: %s" % str(ArbitraryJobAds))
-    Arbirtray_adlist = ArbitraryJobAds.split(",")
-    DebugPrint(5, "Going through list: %s " % Arbirtray_adlist)
-    for arbitaryAd in  Arbirtray_adlist:
-        if arbitaryAd in classad:
-            DebugPrint(5, "Arbitrary ad: %s found in classAd" % arbitaryAd) 
-            DebugPrint(5, "Value of arbitrary classAd: %s" % str(classad[str(arbitaryAd)]))
-            r.AdditionalInfo(arbitaryAd, str(classad[str(arbitaryAd)]))
+    ArbitraryJobAttrs = GratiaCore.Config.getConfigAttribute("ExtraAttributes")
+    DebugPrint(5, "ArbitraryJobAd: %s" % str(ArbitraryJobAttrs))
+    ArbitraryAttrslist = re.split(",?\s?", ArbitraryJobAttrs)
+    #ArbitraryAttrslist = ArbitraryJobAds.split(",")
+    for arbitraryAttr in  ArbitraryAttrslist:
+        if arbitraryAttr in classad:
+            DebugPrint(5, "Arbitrary ad: %s found in classAd, value %s" % (arbitraryAttr, str(classad[str(arbitraryAttr)]))) 
+            r.AdditionalInfo(arbitraryAttr, str(classad[str(arbitraryAttr)]))
     ########################################################################################
     if 'MachineAttrCpus0' in classad:
         r.Processors(classad['MachineAttrCpus0'], metric="max")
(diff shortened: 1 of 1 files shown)


condor/condor_meter
```diff
@@ -711,7 +710,19 @@ def classadToJUR(classad):
 
     setIfExists(r.Queue, classad, "JobUniverse", "Condor's JobUniverse field", setstr=True)
     setIfExists(r.NodeCount, classad, 'MaxHosts', "max")
-
+    ########################################################################################
+    ########################################################################################
+    # Code added to send to arbitrary Ads SOFTWARE-2714
+    ArbitraryJobAds = GratiaCore.Config.getConfigAttribute("ArbitraryJobAdAttributes")
+    DebugPrint(5, "ArbitraryJobAd: %s" % str(ArbitraryJobAds))
+    Arbirtray_adlist = ArbitraryJobAds.split(",")
```
bbockelm: This will not handle space-separated lists (`foo bar baz` or `foo, bar, baz`) as is traditional in HTCondor.

Try:
```
import re
...
Arbitrary_adlist = re.split(",?\s?", ArbitraryJobAds)
```

condor/condor_meter
```diff
@@ -711,7 +710,19 @@ def classadToJUR(classad):
...
+    DebugPrint(5, "Going through list: %s " % Arbirtray_adlist)
+    for arbitaryAd in  Arbirtray_adlist:
```
bbockelm: Variable name `Arbirtray_adlist` is misspelled.  There's also a mix of CamelCase and underscores - looking at the nearby code, stick with CamelCase.

condor/condor_meter
```diff
@@ -711,7 +710,19 @@ def classadToJUR(classad):
...
+        if arbitaryAd in classad:
```
bbockelm: Might consider renaming this to `arbitraryAttr` instead of `arbitraryAd` since it isn't actually an ad...

condor/condor_meter
```diff
@@ -711,7 +710,19 @@ def classadToJUR(classad):
...
+            DebugPrint(5, "Arbitrary ad: %s found in classAd" % arbitaryAd) 
+            DebugPrint(5, "Value of arbitrary classAd: %s" % str(classad[str(arbitaryAd)]))
+            r.AdditionalInfo(arbitaryAd, str(classad[str(arbitaryAd)]))
```
bbockelm: For complex expressions, this will simply convert the expression to a string instead of evaluate the expression.

I think we want to evaluate the expression.

condor/condor_meter
```diff
@@ -711,7 +710,19 @@ def classadToJUR(classad):
...
+    ########################################################################################
+    # Code added to send to arbitrary Ads SOFTWARE-2714
+    ArbitraryJobAds = GratiaCore.Config.getConfigAttribute("ArbitraryJobAdAttributes")
+    DebugPrint(5, "ArbitraryJobAd: %s" % str(ArbitraryJobAds))
+    Arbirtray_adlist = ArbitraryJobAds.split(",")
+    DebugPrint(5, "Going through list: %s " % Arbirtray_adlist)
+    for arbitaryAd in  Arbirtray_adlist:
+        if arbitaryAd in classad:
+            DebugPrint(5, "Arbitrary ad: %s found in classAd" % arbitaryAd) 
+            DebugPrint(5, "Value of arbitrary classAd: %s" % str(classad[str(arbitaryAd)]))
```
bbockelm: Convert the two lines of debug statements into a single one.

condor/condor_meter
```diff
@@ -711,7 +710,19 @@ def classadToJUR(classad):
 
     setIfExists(r.Queue, classad, "JobUniverse", "Condor's JobUniverse field", setstr=True)
     setIfExists(r.NodeCount, classad, 'MaxHosts', "max")
-
+    ########################################################################################
+    ########################################################################################
+    # Code added to send to arbitrary Ads SOFTWARE-2714
+    ArbitraryJobAds = GratiaCore.Config.getConfigAttribute("ArbitraryJobAdAttributes")
```
bbockelm: The config variable name is a mouthful.  How about `ExtraAttributes`?

condor/condor_meter
osg-cat: Won’t that allow it to split every single character apart? Perhaps something like this:

`Arbitrary_adlist = re.split(r'[,\s]+', ArbitraryJobAds)`

bbockelm: Those are equivalent.

osg-cat: They are not equivalent regular expressions, but I guess that they behave similarly in this particular case due to a oddity (IMO) of Python. The docs for re.split() say, “Note that split will never split a string on an empty pattern match”, which is key here.

efajardo:
@bbockelm,

I did the changes you requested.

For testing purposes I modify this file:

`/etc/gratia/condor/ProbeConfig`

With the following:

```Title2="Probe information and functional configuration"
ProbeName="condor:uclhctest.t2.ucsd.edu"
SiteName="UCSD"
Grid="OSG"
EnableProbe="1"
...
ExtraAttributes="foo,foo2,foo3"
```

I then submitted a job with the following submit file

```
output = myoutput.$(Cluster)
error = myerror.$(Cluster)
log = mylog.$(Cluster)
executable = echoScript.sh
+foo = "bar"
+foo2 = "bar2"
+foo3 = "bar3"
queue 10
```

And checked here in [GRAC](https://gracc.opensciencegrid.org/kibana/app/kibana?#/discover?_g=(refreshInterval:(%27$$hashKey%27:%27object:267%27,display:%275%20seconds%27,pause:!t,section:1,value:5000),time:(from:now%2Fw,mode:quick,to:now%2Fw))&_a=(columns:!(ProbeName),index:%27gracc.osg.raw-*%27,interval:auto,query:(query_string:(analyze_wildcard:!f,query:%27ProbeName:%22condor:uclhctest.t2.ucsd.edu%22%27)),sort:!(EndTime,desc))), that the records have:

```? Resource_foo	   	  bar
? Resource_foo2	   	  bar2
? Resource_foo3	   	  bar3
```

Is this allright?

condor/condor_meter
```diff
@@ -711,7 +710,18 @@ def classadToJUR(classad):
...
     setIfExists(r.Queue, classad, "JobUniverse", "Condor's JobUniverse field", setstr=True)
     setIfExists(r.NodeCount, classad, 'MaxHosts', "max")
-
+    ########################################################################################
+    ########################################################################################
+    # Code added to send to arbitrary Ads SOFTWARE-2714
+    ArbitraryJobAttrs = GratiaCore.Config.getConfigAttribute("ExtraAttributes")
+    DebugPrint(5, "Arbitrary Job Attributes: %s" % str(ArbitraryJobAttrs))
+    ArbitraryAttrslist = re.split(",?\s?", ArbitraryJobAttrs)
+    #ArbitraryAttrslist = ArbitraryJobAds.split(",")
```
bbockelm: Please remove commented-out line.

condor/condor_meter
```diff
@@ -770,6 +780,7 @@ def classadToJUR(classad):
         if isinstance(evaluated, types.StringType):
             r.SiteName(evaluated)
             r.ProbeName('%s-%s' % (r.GetProbeName(), evaluated))
+
```
bbockelm: Please remove unnecessary whitespace change.

PR #19, opened 2017-06-29
drop bdii-status probe (SOFTWARE-2660)

edquist:


matyasselmeci:
Does it work?

edquist:
We're just removing it... Nothing else in the gratia probes references bdii anything.  ... I think it should work, but if we want to extra convince ourselves I can make a build and run osg-test against it to make sure nothing starts breaking.  (The lazy side of me wouldn't care enough to do this though.)

PR #22, opened 2017-11-20
Add AWS VM probe

retzkek:
This probe was sitting in a dev branch for a while, it's been unchanged and running for over a year, so we'd like to add it to master and get it into the standard build & release process.

I've done a test RPM build and install on SL7.

### DIFF
--- a/awsvm/ProbeConfig.add
+++ b/awsvm/ProbeConfig.add
@@ -1,0 +1,7 @@
+    ExemptDuplicates="0"
+    DuplicateFilelock="/var/lib/gratia/filelock"
+    HardwareDetailsURL=""
+    HardwareDetailsFile="/usr/share/gratia/awsvm/aws_demand_data_20151102.json"
+    AWSCredentialsFile="~/.aws/credentials"
+    AWSRegions="us-east-1,us-west-1,us-west-2"
+        Comments91="Leave blank to obtain complete region list from AWS"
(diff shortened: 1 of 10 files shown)


retzkek:
Oh, alright then.. the commit history was pretty noisy so I thought a squash would be better. I'm actually now not sure about the `python2-boto3` dependency; I'm told we're installing boto3 with pip anyways for other packages, so there may be a conflict there (I'll find out in the next couple days hopefully). Also, I see it's not available for el6, is that a problem?

edquist:
As for `python2-boto3` not being available for el6 - it means that yum will fail/refuse to install the probe on el6, even if perhaps they have made boto3 available with pip. ... So... I guess I see the options being:
1. drop the `python2-boto3` requirement altogether and expect people to install via pip
2. make the `python2-boto3` requirement conditional for el7 (`%if 0%{?rhel} >= 7 `), and expect people to install via pip on el6
3. convince @brianhlin that we should package `python2-boto3` in osg for el6  ;)
4. expect/accept that nobody will use this probe on el6

(To me, 1 or 2 makes the most sense, but I'm not sure what the expected usage is like...)

... As for the merge vs squash thing, in general I think it's nice to have the history context to see how changes were introduced if anything needs to be investigated later, and for git blame to tell the right story. Plus then the dev branch shows up as merged. ... But if you feel strongly about discarding the history I won't put up any further resistance  ;)

brianhlin:
I would suggest \#4: only package this for EL7 and be done with it

edquist:
@retzkek, I think you were going to see whether you all wanted to keep the boto3 as an rpm requirement vs doing the pip install.  Anyway feel free to merge this as is, or do a non-squash merge of the `aws` branch to preserve the history.

PR #24, opened 2018-01-18
Classad attr are case insensitive (SOFTWARE-3017)

brianhlin:
Accessing classad attr via 'classad.get(attr)' or 'classad[attr]' allows
for case insensitive access but 'attr in classad' does not. This
caused us to miss picking up attributes in the ad

### DIFF
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -23,11 +23,8 @@ import gratia.common.Gratia as Gratia
 import gratia.common.file_utils as file_utils
 import gratia.common.config as config
 
-g_has_classad = True
-try:
-    import classad as classadLib
-except ImportError:
-    g_has_classad = False
+import classad as classadLib
+
 g_alternate_records = {}
 g_probe_config = None
 
@@ -41,83 +38,6 @@ min_start_time = time.time() - 120*86400
 
 class IgnoreClassadException(Exception):
     pass
-
-
-# ********************************************
-# class     caselessDict
-#           http://code.activestate.com/recipes/66315-case-insensitive-dictionary/
-# purpose   emulate a normal Python dictionary
-#           but with keys which can accept the
-#           lower() method (typically strings).
-#           Accesses to the dictionary are
-#           case-insensitive but keys returned
-#           from the dictionary are always in
-#           the original case.
-# ********************************************
-
-class caselessDict:
-    def __init__(self, inDict=None):
-        """Constructor: takes conventional dictionary
-           as input (or nothing)"""
-        self.dict = {}
-        if inDict != None:
-            for key in inDict:
-                k = key.lower()
-                self.dict[k] = (key, inDict[key])
-        self.keyList = self.dict.keys()
-        self.iterPosition = 0
-        return
-
-    def __iter__(self):
-        self.iterPosition = 0
-        return self
-
-    def next(self):
-        if self.iterPosition >= len(self.keyList):
-            raise StopIteration
-        x = self.dict[self.keyList[self.iterPosition]][0]
-        self.iterPosition += 1
-        return x
-
-    def __getitem__(self, key):
-        k = key.lower()
-        return self.dict[k][1]
-
-    def __setitem__(self, key, value):
-        k = key.lower()
-        self.dict[k] = (key, value)
-        self.keyList = self.dict.keys()
-
-    def has_key(self, key):
-        k = key.lower()
-        return k in self.keyList
-
-    def __len__(self):
-        return len(self.dict)
-
-    def get(self, key, alt):
-        if self.has_key(key):
-            return self.__getitem__(key)
-        return alt
-
-    def keys(self):
-        return [v[0] for v in self.dict.values()]
-
-    def values(self):
-        return [v[1] for v in self.dict.values()]
-
-    def items(self):
-        return self.dict.values()
-
-    def __contains__(self, item):
-        return self.dict.has_key(item.lower())
-
-    def __repr__(self):
-        items = ", ".join([("%r: %r" % (k, v)) for k, v in self.items()])
-        return "{%s}" % items
-
-    def __str__(self):
-        return repr(self)
 
 
 # --- functions -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


edquist:
Fascinating.  I feel like the fundamental problem is that `attr in classad` should return true in a case-insensitive fashion if `classad.get(attr)` is case-insensitive -- perhaps that should be feature request for the maintainers of the classad library?  (thoughts, @bbockelm?)

But otherwise, @brianhlin, what you have makes sense.

brianhlin:
This will still work for sites without condor-python libs because the classad is stored as a [caselessDict](https://github.com/opensciencegrid/gratia-probe/blob/master/condor/condor_meter#L58) in that case.

Fixing this at the condor level will work as well but we'll get faster turnaround this way and this method will still work even if they implement case-insensitive `__contains__`

edquist:
@brianhlin - I'm confused.  If the `classad` object is a `caselessDict` which implements a case-insensitive `__contains__`, none of your changes to `condor_meter` should be necessary ... right?

brianhlin:
@edquist - if the `classad` lib is available, the classads are stored as a `classad` object, which doesn't have case-insenstive `__contains__`.

brianhlin:
@djw8605 are there specific reasons why sites wouldn't have condor-python installed?

djw8605:
No specific reasons.  Just the normal ones.  "Old installations", "condor installed not with RPM"...

edquist:
Ah, so it only fails if condor-python is installed  :-p

Now i get it.

brianhlin:
https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6535

Discussing this with @matyasselmeci, let's get rid of `caselessDict` and just require `condor-python`. We already require `condor` and we have `empty-condor-python` for sites that don't install via RPM.

brianhlin:
@edquist I don't have enough cycles to complete this work so I'd like you to handle removal of `caselessDict` for the 3.4.8 release.

edquist:
@brianhlin, this needs to be coordinated with the corresponding [htcondor ticket](https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6535), right?

That is, requiring `condor-python` will only make this work for a version of condor that includes the work to make `__contains__` case-insensitive, right?

Anyway, I can rip out the `caselessDict` stuff for this PR, but my main question/concern is about when it'll be safe to release these gratia-probe changes.

brianhlin:
@edquist nope, it doesn't have to be coordinated with that ticket because I've changed the membership check from using `if x in classad` to `classad[x]`. The former is currently broken, the latter is not.

edquist:
Ohhhhhhhh........

edquist:
I have ripped out the `caselessDict` stuff, if anyone would like to review.

Some thoughts:

the `condor` rpm has already pulled in `condor-python` for some time now, so adding the `condor-python` requirement is somewhat redundant.  But at least it documents that we need it explicitly.

> No specific reasons. Just the normal ones. "Old installations", "condor installed not with RPM"...

Hopefully this new requirement will not break old installations too badly...  (If they are updating to the new gratia-probe, they'll probably be updating to newer condor too?)

> I've changed the membership check from using `if x in classad` to `classad[x]`

Do we need to be concerned about the cases where `classad[x]` is `0` or `""` ?

condor/condor_meter
```diff
@@ -811,66 +733,25 @@ classad_double_re = re.compile("^(\w{1,255}) = ([+-]? *(?:\d{1,30}\.?\d{0,30}|\.
...
+            buffer = ''
         else:
-            yield add_unique_id(classadLib.parseOld(buffer))
+            buffer += lineOrig
+
+    if hasattr(classadLib, 'parseOne'):
+        yield add_unique_id(classadLib.parseOne(buffer))
     else:
-        yield add_unique_id(classad)
+        yield add_unique_id(classadLib.parseOld(buffer))
```
brianhlin: We should be able to get rid of the check for `parseOne` availability and fallback to `parseOld`. `parseOne` has been available in condor since 8.3 (http://research.cs.wisc.edu/htcondor/manual/v8.3/6_7Python_Bindings.html).

edquist:
> Do we need to be concerned about the cases where `classad[x]` is `0` or `""` ?

Specifically, I'm wondering if the check ought to be `classad.get(x) is not None` instead.

brianhlin:
> Hopefully this new requirement will not break old installations too badly... (If they are updating to the new gratia-probe, they'll probably be updating to newer condor too?)

With the `condor-python` dependency, we should be in the clear for these cases. If it breaks something, the site would have to be running some sort of unsupported frankenstein of condor 8.2 but the newest gratia-probe

> Do we need to be concerned about the cases where `classad[x]` is `0` or `""`?

That's a good point, could you update to `classad.get(x) is not None` like you suggested.

condor/condor_meter
```diff
@@ -561,7 +561,7 @@ def determine_host_description(classad):
         return None
 
     # Check first for SE-based matching.
-    if ('DESIRED_SEs' not in classad) or ('MATCH_GLIDEIN_SEs' not in classad):
+    if not classad.get('DESIRED_SEs') or classad.get('MATCH_GLIDEIN_SEs'):
```
edquist: I'm gonna switch this to check against the `None` value - but I think order of operations wise this change is missing a `not` on the right hand of the `or`...

edquist:
Get your hate ready!

I had this fantastic idea.

```
import classad as classadLib

classadLib.ClassAd.__contains__ = lambda self,attr: self.get(attr) is not None
```

Then all the `attr in ad` tests can just remain as they were...!

(@matyasselmeci said he was sorry for not hating on me for this idea.)

brianhlin:
I don't hate it as long as there are explanatory comments with links to the JIRA ticket/PR to go alongside it

edquist:
Well, wow ka-pow, in that case.

edquist:
If you'd like I can rebase this PR to simplify the commits to be merged.

edquist:
Yeah, rebasing (old tip for this branch's history was a870f1f)

brianhlin:
> Pull request authors cannot approve their own pull request

I guess that makes sense. Anywho, LGTM.

brianhlin:
Last call for changes before I merge this

edquist:
What does github have against autocracy?  :-p

I don't know whether to be shocked or tickled that nobody objected to 68d42eb, but it certainly simplified things.

bbockelm:
Ouch.  68d42eb wounds the soul.

Could we make it so the monkey patch is automatically "undone" in future versions?  Do we know when this will be fixed?

brianhlin:
@bbockelm I brought it up in the OSG-FW meeting and Todd added it to his plate since he's already working on Python stuff. No ETA but we'll track it through that meeting and get rid of the hack when it makes it into a condor release.

PR #27, opened 2018-02-08
slurm bug fixes (SOFTWARE-2795)

edquist:
Fixes for bugs reported by Bockjoo Kim in:
https://jira.opensciencegrid.org/browse/SOFTWARE-2795

### DIFF
--- a/slurm/SlurmProbe.py
+++ b/slurm/SlurmProbe.py
@@ -127,19 +127,17 @@ class SlurmProbe:
         return fp.readline().rstrip('\n')
 
     def get_slurm_version(self):
-        prog = "srun --version"
+        prog = "srun"
         path = Gratia.Config.getConfigAttribute("SlurmLocation")
         fd = None
-
-        # Look for the program on the $PATH
-        cmd = prog
 
         # Unless there is a specific path configured
         if path:
             c = os.path.join(path, "bin", prog)
             if os.path.exists(c):
-                cmd = c
+                prog = c
 
+        cmd = "'%s' --version" % prog.replace("'","'\\''")
         fd = os.popen(cmd)
         output = fd.read()
         if fd.close():
--- a/slurm/SlurmProbe.py
+++ b/slurm/SlurmProbe.py
@@ -15,6 +15,7 @@
 import sys, os, stat
 import time, random
 import pwd, grp
+import subprocess
 
 from gratia.common.Gratia import DebugPrint
 import gratia.common.GratiaWrapper as GratiaWrapper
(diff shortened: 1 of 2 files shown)


slurm/SlurmProbe.py
```diff
@@ -140,7 +140,7 @@ def get_slurm_version(self):
             if os.path.exists(c):
                 cmd = c
 
-        fd = os.popen(prog)
+        fd = os.popen(cmd)
```
djw8605: This breaks the path `SlurmLocation` configuration.  `cmd` is just the `srun --version`, and `prog` should be the resolved version after accounting for `SlurmLocation`

edquist: Thanks @djw8605!

I totally missed the second problem, that the `SlurmLocation` lookup would never succeed, since it was testing for the existence of `prog` which was set to `srun --version`.

Let me know if this fixes it.

djw8605:
Now I'm more confused.  Why are we doing some odd `replace` call?  Why not just prepend the path in `SlurmLocation`? (like before).  I guess I don't understand why it didn't work before?

edquist:
Aha, fair enough -- i'll step through and explain what i'm seeing.

The relevant section in the current `master` (663eac2633f7b58d6579c87a40513d295fa12254) is:

https://github.com/opensciencegrid/gratia-probe/blob/663eac2633f7b58d6579c87a40513d295fa12254/slurm/SlurmProbe.py#L130-L143

> I guess I don't understand why it didn't work before?

Note that `prog` is set to `"srun --version"`, and thus the `os.path.exists()` check effectively tests whether the path `"$SlurmLocation/bin/srun --version"` exists, which presumably never worked.  (That is, the `os.path.exists()` check presumably never worked, so `SlurmLocation` never got used.)

My fix is to set `prog` to just `"srun"`, in order to properly do the existence check, and then append `--version` to the `cmd` that gets run.

> Why are we doing some odd replace call? Why not just prepend the path in SlurmLocation? (like before).

If safe shell is hard to come by, when nobody actually wants it ;)

The reason is because `os.popen()` takes a single `str` argument for the cmdline, which is interpreted by the shell (it does not accept a list).  This was fine before, since the `prog` cmdline was always `srun --version` (no problematic shell metacharacters).  But prepending an arbitrary (user-input controlled) path will cause the shell to wrongly interpret any spaces or other metacharacters in the filename.

Doing `"'%s'" % arg.replace("'","'\\''")` is a cheap but (i believe) correct way to escape an arbitrary command line argument for the shell.

If you'd prefer, i can shove this into a helper function for clarity, or use `subprocess.Popen` (which accepts a list of args for a cmdline) instead of `os.popen`.

brianhlin:
Isn't `subprocess.Popen()` the preferred function in Python2?

edquist:
> Isn't `subprocess.Popen()` the preferred function in Python2?

`os.popen()` still has a bit simpler interface and in any case it's what the probe was using already.  But i did not realize the docs label it as deprecated now in favor of `subprocess.Popen()`, so, yeah i suppose that's the recommended replacement now.

brianhlin:
Let's shy away from deprecated functions where we can :)

edquist:
Normally i shy away from changing more than i need to, but in any case i've rejiggered it now to use `subprocess.Popen` now instead of the cozy old `os.popen`.

PR #33, opened 2018-06-06
Use condor_ce_config_val on htcondor-ce (SOFTWARE-2629)

sthapa:


condor/condor_meter
```diff
@@ -774,7 +774,11 @@ def htcondor_configured():
     """ 
     try:
         path = GratiaCore.Config.getConfigAttribute("CondorLocation")
-        condor_config_path = os.path.join(path, "bin", "condor_config_val")
+        if 'htcondor-ce' in g_probe_config and g_probe_config is not None:
```
brianhlin: Can you change this to `g_probe_config.get('htcondor-ce')`?

condor/condor_meter
```diff
@@ -774,7 +774,11 @@ def htcondor_configured():
...
+          condor_binary = 'condor_ce_config_val'
```
brianhlin: `condor_config_binary` would be more descriptive

condor/condor_meter
brianhlin: Scratch that, I misread the statement and thankfully @edquist had a strong opinion about my suggestion. The statement is still broken though, it should read:

`if g_probe_config is not None and 'htcondor-ce' in g_probe_config:`

sthapa: Fixed

condor/condor_meter
sthapa: renamed

PR #34, opened 2018-06-19
attempt to update last mod time for empty log files (SOFTWARE-3041)

edquist:
"empty" as in we didn't process any log events from it and we weren't
otherwise planning to skip it (because it was the most recent).

This is true for a log with size 0, or if (for example) it contains only
empty lines.

### DIFF
--- a/pbs-lsf/urCollector-src/urCollector.pl
+++ b/pbs-lsf/urCollector-src/urCollector.pl
@@ -375,7 +375,7 @@ sub processLrmsLogs {
          print "".localtime().": Processing LRMS log file: $thisLogFile; last modified: $logFMod{$thisLogFile}\n";
          
          &processLrmsLogFile($thisLogFile, $newestFile,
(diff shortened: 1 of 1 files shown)


edquist:
Heya @djw8605, I don't know how familiar you are with the big perl script in the pbs-lsf probe, but if you're comfortable maybe you can give it a sanity check...

The idea here is to update the last timestamp even when log files are empty, where previously they would only get updated based on the event timestamps in the (non-empty) logs.

Of course, I'd be open to suggestions if you think there's a better way to handle it.

edquist:
Haha, yeah, it's fun to hate on this probe.

It took me some time to realize that `$_[5]` was actually being used to pass the last timestamp back to the caller (because perl subroutine args are aliases).

(Unrelated to this PR specifically, what really gets to me about the pbs-lsf probe is that it still has 2.7K lines of C++ whose only purpose is to generate 30 lines of xml...)

PR #35, opened 2018-06-26
Added a few high-level Pegasus attributes for collection under the HT…

rynge:
…Condor probe

This is related to a request from Miron/Ewa on improving classification of user workloads. With these attributes we will be able to group related jobs together (at least for the Pegasus case), and get some idea on what tools are in use.


brianhlin:
@edquist @djw8605 should we also add this to the HTCondor-CE probe?

edquist:
@brianhlin, any changes to the condor probe are also changes to the htcondor-ce probe

PR #36, opened 2018-07-23
don't consider jobs completed within the last 15 min (SOFTWARE-3347)

edquist:
The idea here is to introduce a delay to account for a potential lag in jobs getting logged.

### DIFF
--- a/slurm/slurm_meter
+++ b/slurm/slurm_meter
@@ -15,6 +15,7 @@
 from SlurmProbe import SlurmProbe
 from gratia.common.Gratia import DebugPrint
 
+import time
 import optparse
 import gratia.common.Gratia as Gratia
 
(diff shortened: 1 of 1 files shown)


slurm/slurm_meter
```diff
@@ -214,6 +215,9 @@ class SlurmMeter(SlurmProbe):
         # Loop over completed jobs
         time_end = None
         server_id = self.get_db_server_id()
+        # don't consider jobs that have ended within the last 15 min
+        # https://opensciencegrid.atlassian.net/browse/SOFTWARE-3347
+        cutoff_ts = min(self.checkpoint.val, int(time.time()) - 15 * 60)
         for job in self.sacct.completed_jobs(self.checkpoint.val):
```
djw8605: Don't you have to send `cutoff_ts` to the complete_jobs function?

edquist: Yeah derp.  I realized this last night in my sleep :D

djw8605:
Looks good to me.  I'll let you merge.  Can you send me a build for bockjoo to test (and I'll test at Nebraska too).

edquist:
Sounds good, will do.

PR #38, opened 2018-08-21
log unhandled exceptions in slurm probe (SOFTWARE-2456)

edquist:
the traceback still goes to stderr, but never got logged to the gratia
logs.  by using DebugPrint, at least the error will make it to the
gratia logs.

### DIFF
--- a/build/gratia-probe.spec
+++ b/build/gratia-probe.spec
@@ -1,7 +1,7 @@
 Name:               gratia-probe
 Summary:            Gratia OSG accounting system probes
 Group:              Applications/System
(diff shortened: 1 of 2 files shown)


edquist:
Nyope...

edquist:
A couple tweaks: the log line didn't stand out much, so i added a more prominent ERROR: prefix to the line, and I changed raise e to simply raise since the latter is necessary to preserve the original traceback.

I tested the new code by running the probe with bad parameters in the ProbeConfig, and verified that the unhandled exceptions get logged to the gratia logs as expected.

PR #39, opened 2018-09-07
cast classad __contains__ key check to str (SOFTWARE-3415)

edquist:
Patch provided by Brian B


edquist:
See also #40, which i think i prefer over the solution here.

PR #43, opened 2019-01-09
Processors field: Evaluate ClassAd expr and ensure a default of 1 (SOFTWARE-3474, SOFTWARE-3476)

brianhlin:
@djw8605 @edquist this is a WIP branch that modifies the logic for setting the `Processor` field for records coming from the condor probe. The big things are:

1. ClassAd expressions are evaluated to address [SOFTWARE-3474](https://opensciencegrid.atlassian.net/browse/SOFTWARE-3474)
1. Ensures that `Processors` is set to 1 if the ClassAd attributes don't return an integer when evaluated

I'm marking this as `do-not-merge` and holding off on requesting a review because I'm still working through Travis issues and I reserve the right to squash and force-push to my branch at any point.

However, I wanted to get a conversation going about setting the default CPUs to 1. I believe this was the intention of the previous logic but it may serve to obfuscate issues in the future by creating subtly broken records. That however, may be preferable to the current behavior of creating completely broken records, which is the case for hosted CEs. @osg-cat may have some opinions on this -- maybe we can set a `ProcessorsError` field to an informative for records where the expected attributes do not evaluate properly?

### DIFF
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -39,6 +39,9 @@ max_batch_size = 500
 
 min_start_time = time.time() - 120*86400
 
+# The preferred order of JobAd attributes to use for determining the number of
+# processors available to the pilot or payload job
+PROC_ATTRS = ['MachineAttrCpus0', 'MATCH_EXP_JOB_GLIDEIN_Cpus', 'RequestCpus']
 
 # --- classes -------------------------------------------------------------------------
 
@@ -160,9 +163,10 @@ def get_num_procs(job_ad):
     If none of the above are set or do not evaluate to an integer, return 0
     """
     procs = 0
-    for attr in ['RequestCpus', 'MATCH_EXP_JOB_GLIDEIN_Cpus', 'MachineAttrCpus0']:
+    for attr in PROC_ATTRS:
         try:
             procs = int(job_ad.eval(attr))
+            break
         except (KeyError, ValueError):
             continue
 
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -157,17 +157,14 @@ def get_num_procs(job_ad):
     2. MATCH_EXP_JOB_GLIDEIN_Cpus
     3. RequestCpus
 
-    If none of the above are set or do not evaluate to an integer, return 1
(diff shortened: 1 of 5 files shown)


brianhlin:
@djw8605 I think this is more or less ready for review, with one caveat: we don't want to set a default value of 1 (and tbh, we should use the actual number of CPUs instead of falling back to `RequestCpus`). If none of the attributes evaluate properly, 

1. What should we set `Processors` to in the record so that OSG staff can find these bad records in the GRACC
1. What information should we provide to ease troubleshooting?
1. Where do we put that information?

edquist:
>we don't want to set a default value of 1

I think other probes (eg `pbs-lsf`) have elaborate logic to determine processors, but in the end still fall back to 1.  Not sure how best to indicate this condition in GRACC, maybe a new attribute?

>we should use the actual number of CPUs instead of falling back to `RequestCpus`

"actual" as in CPUs on the machine? or the slot?  What meaning of `Processors` is GRACC (or GRACC users) expecting?

brianhlin:
>"actual" as in CPUs on the machine? or the slot? What meaning of `Processors` is GRACC (or GRACC users) expecting?

Since jobs are often running opportunistically at sites, we want to report how many procs the job has available to it, i.e. the slot.

brianhlin:
I had a discussion with @djw8605 and we settled on a default value of 0. Updated the PR accordingly.

condor/condor_meter
```diff
@@ -160,9 +163,10 @@ def get_num_procs(job_ad):
     If none of the above are set or do not evaluate to an integer, return 0
     """
     procs = 0
-    for attr in ['RequestCpus', 'MATCH_EXP_JOB_GLIDEIN_Cpus', 'MachineAttrCpus0']:
+    for attr in PROC_ATTRS:
```
edquist: +1 for this (checking the attrs in order and returning after the first match_ - i was going to suggest you do something similar

brianhlin: Yeah, the original way felt dirty. If there's nothing else, can you approve the PR?

djw8605:
I enabled gratia-probe tests in travis-ci.

PR #44, opened 2019-01-15
touch condor/__init__.py for travis unit tests (SOFTWARE-3530)

edquist:


edquist:
Thanks @brianhlin.

Just to make a note it here to clarify for later development; the layout for python files (`.py`) to be installed on the system is: the subpackage directory, followed by the python path root (ie, site-packages).

Thus, `common/gratia/common/__init__.py`
gets mapped to the `gratia-probe-common` rpm, installed as:
`/usr/lib/python2.6/site-packages/gratia/common/__init__.py`

The `condor` subpackage, on the other hand, does not install any `.py` files, thus it does not have this python path structure, nor an `__init__.py`.

PR #46, opened 2019-04-16
add readme detailing gracc fields for slurm record (SOFTWARE-3641)

edquist:


slurm/README.gracc-fields
```diff
@@ -0,0 +1,78 @@
...
+        <StartTime urwg:description="Was entered in seconds">2019-03-26T19:17:32Z</StartTime>
+        <Queue>hpg1-compute</Queue>
+        <ProjectName>avery</ProjectName>
+        <Memory urwg:description="RSS" urwg:metric="total" urwg:storageUnit="kB">20480</Memory>
+        <ProbeName>slurm:cms.rc.ufl.edu</ProbeName>
+        <SiteName>UFlorida-HPC</SiteName>
+        <Grid>OSG</Grid>
+        <Njobs>1</Njobs>
+        <Resource urwg:description="ResourceType">Batch</Resource>
+</JobUsageRecord>
```
brianhlin: The indentation seems off for this XML. And actually, I think this is the more useful information to break down in the README for the gratia probe. What I'm ultimately after is a description of the structure and fields of a gratia probe:

- Each record has info identifying the record, user, and job as well as info regarding job usage
- What are the human descriptions of each field?

retzkek: @brianhlin the GRACC [dev docs](https://opensciencegrid.org/gracc/dev-docs/raw-records/) go into this a bit, mainly though linking to this doc, which goes into far more detail than I'm sure you ever wanted: https://www.ogf.org/documents/GFD.98.pdf

edquist: >The indentation seems off for this XML

I agree but apparently the XML actually looks like that.

>And actually, I think this is the more useful information to break down in the README for the gratia probe

Yes, since that's what's actually sent.  And actually if you just look at the xml, it's mostly self-explanatory.  But sure, we can document some slightly more human readable descriptions for the fields too.


Some links for future reference:

https://github.com/opensciencegrid/gratia-probe/blob/master/common/gratia/common/record.py
https://github.com/opensciencegrid/gratia-probe/blob/master/common/gratia/common/xml_utils.py
https://github.com/edquist/osg-twiki-backup/blob/cleanup/data/Accounting/ProbeDevelopement.txt

edquist: Thanks @retzkek!  I did not know that super useful gracc/dev-docs/raw-records doc existed!

brianhlin: @retzkek Thanks, that's really helpful!

> And actually if you just look at the xml, it's mostly self-explanatory.

Yeah, except when it's not :P. Like when we use the SiteName to actually refer to the topology resource name, or what does project name correspond to?

edquist: >Like when we use the SiteName to actually refer to the topology resource name, or what does project name correspond to?

I think it's just the `SiteName` as defined in the `ProbeConfig`

brianhlin: I'm more interested in the information that we intend to convey in these values rather than their provenance. For instance, if I give you a default `ProbeConfig`, what should you enter into the `SiteName` field?

slurm/README.gracc-fields
```diff
@@ -0,0 +1,78 @@
...
+StartTime                        March 26th 2019, 19:17:32.000
+Status                           0
+VOName                           /cms/uscms/Role=pilot/Capability=NULL
+WallDuration                     1494
+checksum                         ad6d91a9e96da9c6b444381f6d16f1bb
+lag_time                         89.16
+lag_time_hr                      0.025
+type                             JobUsageRecord
+
+
```
brianhlin: Let's remove the GRACC-record stuff since that's more relevant for the GRACC than the gratia-probe repo.

edquist:
@brianhlin, I updated this PR earlier with an xml sample, but I am guessing it did not send an email as it was a force-push.

PR #47, opened 2019-04-18
provide sample xml for condor probe records (SOFTWARE-3609)

edquist:


condor/README-xml.md
```diff
@@ -105,3 +105,60 @@ Here are sample XML records sent for ResourceType = Batch and BatchPilot
 </JobUsageRecord>
 ```
 
+#### Notes
+
+The key difference is the "payload" jobs (ResourceType=Payload in GRACC, ResourceType=BatchPilot in the xml sent by the
+gratia-probe) are identified when they have a MATCH_EXP_JOBGLIDEIN_ResourceName attribute in their classad, as found in
+the condor history logfile.
```
brianhlin: We may as well reference where that configuration comes from: https://github.com/opensciencegrid/gratia-probe/blob/master/condor/99_gratia-gwms.conf

djw8605: The condor probe normally reads from the per job history directory.  Not directly from `condor_history`, though is has the capability to read from `condor_history` (history logfile).

edquist: Should we just say "as found in the condor history" ?

djw8605: Maybe, "As found in the HTCondor per job history file"

edquist:
>Could we just rename this file to README.md? It seems like just a good place to drop dev knowledge on the condor gratia probe in general

Given that we are so far just detailing xml, I felt like the less-generic filename was more appropriate.

I have in mind to save the more general `README[.md]` name for something more all-encompasing (at least for the probe / directory where it's found).  If we wanted a place to drop random dev knowledge i might be inclined to call it `NOTES` or something.

brianhlin:
Well, it contains more than just XML examples. The note that we're discussing describes the difference between payload and pilot jobs and the mechanism that differentiates them.

You can have READMEs per directory so I'm imagining this README can have condor-probe specific developer notes.

PR #48, opened 2019-05-10
parse sge records into namedtuple instead of dict

edquist:
doesn't fix any bugs, but speeds up parsing by ~10x !

Note that there are curiously two extra fields at the end that were previously not listed in `sge_meter`.  In the field, i've seen these columns to always have the value `0`, and I haven't looked into what they are for.  Since I don't know and since the probe doesn't use them, I've uninterestingly called them `unknown1` and `unknown2`.

### DIFF
--- a/sge/sge_meter
+++ b/sge/sge_meter
@@ -27,7 +27,7 @@
 
 
 import gratia.common.Gratia as Gratia
-import os, copy, pwd, sys, string
+import os, copy, pwd, sys, string, collections
 
 # This module requires python v2.3
 from optparse import OptionParser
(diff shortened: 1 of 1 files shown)


brianhlin:
Interesting, so the unknown fields are from SGE itself and we just ignored them in the past?

edquist:
Correct.  One detail

djw8605:
According to the documentation: https://linux.die.net/man/5/sge_accounting

They are:
> **arid**
> Advance reservation identifier. If the job used resources of an advance reservation then this field contains a positive integer identifier otherwise the value is "0" .

> **ar_submission_time**
> If the job used resources of an advance reservation then this field contains the submission time (GMT unix time stamp) of the advance reservation, otherwise the value is "0" .

edquist:
Ah, thanks @djw8605, that explains it then.

brianhlin:
@edquist is there anything else that would prevent this from being merged?

edquist:
@brianhlin, I don't have anything else left to add for this PR.  You and/or @djw8605 can feel free to give it a review.

Note that there is nothing important about getting this change in, it's just a performance enhancement.  (Which came up for me when i was re-parsing a large set of old records.)

edquist:
Alrighty then!  :)

PR #49, opened 2019-06-07
support slurm v18 schema (SOFTWARE-3724)

edquist:
The slurm db schema changes in v18; apparently the max_rss field goes
away.

Bockjoo had success replacing the max_rss subquery with
"MAX(j.mem_req)".

We'll use this if the slurm version is 18+, otherwise we'll use the
existing max_rss subquery.

Note that we add slurm_version to SlurmAcctBase, which does not have
access to SlurmProbe.get_slurm_version()

### DIFF
--- a/slurm/SlurmProbe.py
+++ b/slurm/SlurmProbe.py
@@ -87,12 +87,13 @@ class SlurmProbe:
         self.cluster = Gratia.Config.getConfigAttribute('SlurmCluster')
 
         # SLURM made changes to the accounting database schema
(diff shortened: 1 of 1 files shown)


edquist:
Hi @djw8605, would you be able to take a peek at this?

edquist:
Yeah maybe we'll make another change soon for v19 or maybe the htcondor-ce probe will work for Bockjoo.  Patrick was also having issues with v18 though so this may still be worthwhile as-is.

PR #50, opened 2019-07-16
slurm v19 compat update from Bockjoo (SOFTWARE-3740)

edquist:
Alternate sql provided by Bockjoo to work with slurm v19


edquist:
#52 should supersede this PR.  The fix is basically identical but is not limited to slurm v19, since the problem seems to have cropped up due to the new mysql version, not slurm v19.

PR #54, opened 2019-10-30
log unhandled errors (SOFTWARE-3877)

edquist:
As encountered in SOFTWARE-3877, a history file parsing error does not get caught anywhere in the condor probe, and so the exception / traceback does not get logged to the gratia logs.  (condor_meter is run by cron, so any stderr output goes to ..? probably syslog or root email.)

This is a first step to log unexpected failures to the gratia logs.

Beyond this it would be good to individually catch history files that cause parse errors.

### DIFF
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -936,5 +936,13 @@ def send_alternate_records_child(info, record_list):
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except SystemExit:
+        raise
+    except KeyboardInterrupt:
+        raise
+    except Exception, e:
+        DebugPrint(-1, "ERROR: Unexpected error encountered: %s" % e)
+        raise
 
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -307,8 +307,14 @@ def process_history_dirs(dirs):
         # Make sure the filename is in a reasonable format
         m = condor_history_re.match(logfile)
         if m:
-            cnt_submit, cnt_found, cnt_alternate = process_history_file(log)
-            if cnt_submit + cnt_alternate == cnt_found and (cnt_submit > 0 or cnt_alternate > 0):
+            e = None
+            try:
+                cnt_submit, cnt_found, cnt_alternate = process_history_file(log)
+            except ValueError, e:
+                DebugPrint(1, "Failed to parse log file: %s\nError was: %s" % (log, e))
+                cnt_submit, cnt_found, cnt_alternate = 0, 0, 0
+
+            if not e and cnt_submit + cnt_alternate == cnt_found and (cnt_submit > 0 or cnt_alternate > 0):
                 DebugPrint(5, "Processed %i ClassAds from file %s" % (cnt_submit, log))
             else:
                 DebugPrint(2, "Unable to process ClassAd from file (will add to quarantine): %s.  Submit count %d; found count %d" % (log, cnt_submit, cnt_found))
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -944,5 +944,6 @@ if __name__ == "__main__":
         raise
     except Exception, e:
         DebugPrint(-1, "ERROR: Unexpected error encountered: %s" % e)
+        DebugPrintTraceback()
         raise
 


djw8605:
From the log, it appears that the probe continues and connects to the collector to upload the records that were successfully parsed.  Does the probe completely exit, like this pull request indicates?

edquist:
Well, that's an interesting data point.  Did I misread the code?  I mean it looks like when `classadLib.parseOne()` is called [here](https://github.com/opensciencegrid/gratia-probe/blob/1342d3f55d52ef59ce7d747556a55a1c5c153426/condor/condor_meter#L777) (which at least for me raised a `ValueError` on that truncated history file), there's nowhere all the way up the call chain where it gets caught in a `try` block.

So i believe it causes the whole thing to exit immediately, but if you see a way for it not to please point it out!  :)

djw8605:
Agree that I don't know where it is catching the exception.  But, the log shows it must be (from jira):
```
13:06:14 EDT Gratia: DEBUG: After SearchOutstandingRecord outbox:0 staged outbox:0 tarfiles:0
13:06:14 EDT Gratia: Reprocessing response: OK - Reprocessing 0 record(s) uploaded, 0 bundled, 0 failed
13:06:14 EDT Gratia: After reprocessing: 0 in outbox 0 in staged outbox 0 tar files
13:06:14 EDT Gratia: Running command to check condor config: /usr/bin/condor_config_val -schedd PER_JOB_HISTORY_DIR
13:06:14 EDT Gratia: We will process the following directories: /var/lib/gratia/data/.
13:06:14 EDT Gratia: Processing directory /var/lib/gratia/data/.
13:06:14 EDT Gratia: Processing logfile history.2011518.0
13:06:14 EDT Gratia: Processing bundle file: 
13:06:14 EDT Gratia: DEBUG: Connect
13:06:14 EDT Gratia: DEBUG: Connect: OK
13:06:14 EDT Gratia: Connection via HTTP to: gratia-osg-prod.opensciencegrid.org:80
13:06:14 EDT Gratia: DEBUG: Connection SUCCESS```

Immediately after it tries to process the logfile, `history.2011518.0`, it goes right into uploading.

retzkek:
This sounds familiar, I think you're seeing the `atexit` cleanup registered [here](https://github.com/opensciencegrid/gratia-probe/blob/1342d3f55d52ef59ce7d747556a55a1c5c153426/common/gratia/common/GratiaCore.py#L110).

edquist:
Raaaar..... Thanks @retzkek for the tip!  That was mindboggling.

brianhlin:
@edquist what's the status on this PR?

edquist:
Yeah, the [atexit](https://github.com/opensciencegrid/gratia-probe/blob/1342d3f55d52ef59ce7d747556a55a1c5c153426/common/gratia/common/GratiaCore.py#L110) register thing explains why records are getting sent after an uncaught exception is raised.

But it looks like in the actual `__disconnect_at_exit__` function registered with atexit, it only makes it _look_ like there is some exception handling ([here](https://github.com/opensciencegrid/gratia-probe/blob/1342d3f55d52ef59ce7d747556a55a1c5c153426/common/gratia/common/GratiaCore.py#L67-L77)), but in reality that is not for uncaught exceptions from elsewhere in the program, but just for catching exceptions in the exit cleanup code.

So, I think adding the try/except around main() with DebugPrint is still worthwhile, and maybe a DebugPrintTraceback also.

But more to the problem I'll also add some logic to call GratiaCore.QuarantineFile() rather than letting the whole thing die.

edquist:
I've added exception handling to quarantine a log file that produces a `ValueError`.  That's the one we hit anyway.  In theory other errors could be possible while parsing too ... But I'm not sure if it's better to try to handle everything here or just the ones we know are possible.  Other unhandled errors will still be logged to the gratia log.

One thing to note: we hit the error inside of `fd_to_classad`, which is called from both `process_history_file` and `process_history_fd`.  But the latter appears to be used only for reading pipe output from condor_history, and not an actual log file that (1) might be truncated or (2) can be meaningfully quarantined.  So, I've only added this try/except around the call to `process_history_file`, which I believe is the one we hit and in any case the only one with a file that we can quarantine.

@djw8605, let me know what you think of this way of handling things, or if you would suggest going about it differently.

djw8605:
Yup, that all sounds good.

edquist:
Thanks @djw8605 for looking it over..!

Yeah the logic was a bit complicated, as you said, and i tried to change as little as possible to add quarantining at the right spot rather than do a major restructuring... But suggestions welcome if you have an idea for a different approach, feel free to make a PR.

PR #58, opened 2020-05-05
allow non-word characters in cluster name (SOFTWARE-4032)

edquist:
the only invalid characters really are the backtick and a nul byte - the rest should be ok since the cluster name is escaped in backticks when it's used.

### DIFF
--- a/slurm/SlurmProbe.py
+++ b/slurm/SlurmProbe.py
@@ -204,7 +204,7 @@ class SlurmAcctBase(object):
     def __init__(self, conn, cluster, slurm_version):
         self._conn = conn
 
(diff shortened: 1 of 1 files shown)


djw8605:
Hopefully we won't have unicode issues, but otherwise, this is fine.

edquist:
> Hopefully we won't have unicode issues,

Well in any case the raw (utf8) bytes will be preserved (unicode doesn't have any nul characters).  I am not sure how mysql reacts to unicode, but i suppose if they were able to create a table with (non-ascii) unicode characters, then this will still be the correct way to access those tables.

Though who would do such a thing?  ;)

PR #59, opened 2020-06-22
detect condor/htcondor-ce probe config automatically (SOFTWARE-3873)

edquist:
Dave didn't seem to be wild about not being able to just run the htcondor-ce probe directly (without having to specify the `-f` option for the ProbeConfig) ... what would you think about doing it this way instead?

### DIFF
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -37,6 +37,9 @@ g_probe_config = None
 prog_version = "%%%RPMVERSION%%%"
 max_batch_size = 500
 
+probe_name = os.path.basename(os.path.dirname(os.path.abspath(__file__)))
+probe_config = "/etc/gratia/%s/ProbeConfig" % probe_name
+
 min_start_time = time.time() - 120*86400
 
 # The preferred order of JobAd attributes to use for determining the number of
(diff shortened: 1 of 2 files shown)


brianhlin:
This still supports the OSG submit host use of `condor_meter`, right?

edquist:
> This still supports the OSG submit host use of `condor_meter`, right?

Probably..!  How is it used there?  If it's running it with an explicit `-f /path/to/ProbeConfig` then it will continue to work fine.

djw8605:
I think the submit host just uses the normal condor probe.

edquist:
Yeah, apparently @brianhlin in his question was referring to the ProbeConfig being configured for flocking \[1\], though in this case the ProbeConfig still lives in the standard location and the condor probe is still run from the standard location, so nothing about a submit host setup should be affected by this change.

In fact, the only scenario which I can see this causing trouble with would be if you were running a copy of the condor_meter probe, copied to some other directory that was not named `condor`, but still expected it to use the standard `/etc/gratia/condor/ProbeConfig` config path _without_ specifying a `-f` option.

\[1\] https://opensciencegrid.org/docs/submit/osg-flock/#configuring-reporting-via-gratia

PR #62, opened 2020-10-08
add python2/3 compat for pbs-lsf (SOFTWARE-4287)

edquist:
This is the only python piece of pbs-lsf.  Just needed to update the print syntax.


brianhlin (approved):
LGTM

PR #63, opened 2020-10-08
add python2/3 compat for pbs-lsf (SOFTWARE-4287)

edquist:
@matyasselmeci looks like i forgot the "2" part of python2/3 compat in #61


matyasselmeci:
I only see the changes you make, not the changes you don't make ;-)

PR #64, opened 2020-10-09
fix inconsistent indentation for py2/3 compat (SOFTWARE-4287)

edquist:
See them stray tabs?

Those don't fly in python3.


brianhlin (approved):
gratia_khan

PR #66, opened 2020-10-15
add dockerfile for testing probes (SOFTWARE-4313)

edquist:


test/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+       http://repo.opensciencegrid.org/osg/3.5/osg-3.5-el7-release-latest.rpm \
+    && yum install -y --enablerepo=devops osg-build \
+    && yum install -y --enablerepo=osg-empty slurm \
+    && git clone https://github.com/opensciencegrid/gratia-probe \
+    && ver=$(awk '/Version:/ {print $2}' gratia-probe/rpm/gratia-probe.spec) \
+    && yum-builddep -y gratia-probe/rpm/gratia-probe.spec \
+    && mkdir -p bld/upstream \
+    && echo "type=github repo=opensciencegrid/gratia-probe tag=master tarball=gratia-probe-$ver.tar.gz" > bld/upstream/gp.source \
+    && osg-build rpmbuild bld \
+    && yum install -y ./bld/_build_results/gratia-probe-*-*-*.rpm
```
brianhlin, line 19: Can you split this up into different `RUN` invocations? Having them in a single layer means that rebuilding the image will force us to rebuild the entire image instead of only building the layers after the earliest change

test/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM centos:centos7
+
+LABEL name="EL7 OSG 3.5 Gratia Probe Test Container"
+
+RUN cd \
+    && yum install -y yum-priorities \
+    && yum install -y \
+       http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm \
+    && yum install -y \
+       http://repo.opensciencegrid.org/osg/3.5/osg-3.5-el7-release-latest.rpm \
```
brianhlin, line 10: Any reason you're doing this instead of using `software-base`? I can go either way since this way we're not bringing in supervisor

test/Dockerfile
edquist: And here i thought it was an optimization to lump them all together!  :)

Sure, i can split it up.

How do you rebuild just the later layers?  (I don't actually know off hand)

test/Dockerfile
edquist: I actually looked at `opensciencegrid/software-base:fresh` first, but when i tried to docker-run that, it seemed to be "doing something" rather than just giving me a prompt; so i decided to just make a new one.  Otherwise i'm not against using software-base if that can be made to work.

test/Dockerfile
brianhlin: It should do it automatically if you've got a local copy of the image! It's more evident with things like `COPY` later in a Dockerfile. If you build it locally, then change the contents of the file that you're copying in, it'll only rebuiild the COPY + subsequent layers

test/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN yum install -y yum-priorities \
+    && yum install -y \
+       http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm \
+    && yum install -y \
+       http://repo.opensciencegrid.org/osg/3.5/osg-3.5-el7-release-latest.rpm \
+    && yum install -y --enablerepo=devops osg-build \
+    && yum install -y --enablerepo=osg-empty slurm
+
+RUN cd \
+    && git clone --depth=1 https://github.com/opensciencegrid/gratia-probe \
```
brianhlin, line 14: Any reason why you're doing this instead of a `COPY` of the repo? Though idk if you can copy `..`

test/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    && ver=$(awk '/Version:/ {print $2}' gratia-probe/rpm/gratia-probe.spec) \
+    && yum-builddep -y gratia-probe/rpm/gratia-probe.spec \
+    && mkdir -p bld/upstream \
+    && echo type=github repo=opensciencegrid/gratia-probe tag=master \
+            tarball=gratia-probe-$ver.tar.gz > bld/upstream/gp.source \
+    && osg-build rpmbuild bld \
+    && yum install -y ./bld/_build_results/gratia-probe-*-*-*.rpm
```
brianhlin, line 21: Let's do separate `RUN`s for each of these.

brianhlin (commented):
It looks like it'll work as a starting point but I'd like to see default `ProbeConfig`s for testing purposes

test/Dockerfile
brianhlin: The "doing something" is just supervisor (service management for containers) and we don't have to use it here but it may actually be nice to have it just "hang". That way your dev workflow would look liike:

1. `docker run -v <path to host condor_meter>:<wherever condor_meter goes in the container>--name my-probe-dev-env --detach gratia-probe`
1. Make changes to `condor_meter` on your host
1. `docker exec -it my-probe-dev-env condor_meter <options>`
1. See output from above, make additional changes to `condor_meter`, etc.

test/Dockerfile
edquist: Ah, for some reason i thought you wanted whatever was the latest from github.

So do you want to use everything from the checked-out repo that the docker image is built from?

brianhlin: Yeah, I think that makes the most sense as a dev tool!

test/Dockerfile
edquist: Ok... All of them?  :)

I mean i split it up this way because changes to the probes pretty much imply the second half of the steps need to be done also.  So i am curious if there will be a benefit to the extra layers

test/Dockerfile
edquist:
```
COPY failed: Forbidden path outside the build context: .. ()
```

test/Dockerfile
brianhlin: Yeah, I think it makes it a little clearer if a particular command fails

test/Dockerfile
brianhlin: Ha, that's what I suspected. From the root of the repo, does something like the following work?

```
docker build -t gratia-probe -f test/Dockerfile .
```

https://stackoverflow.com/questions/24537340/docker-adding-a-file-from-a-parent-directory/34300129#34300129

test/Dockerfile
edquist: Is there a way to log in and look around at files (logs, quarantine) ?

Or do you have to mount those dirs also?

test/Dockerfile
edquist: Ok, you got it.

edquist:
> It looks like it'll work as a starting point but I'd like to see default `ProbeConfig`s for testing purposes

Ok i'll set those up.

test/Dockerfile
brianhlin: Yup: `docker exec -it my-probe-dev-env /bin/sh`

edquist:
@brianhlin, I've added ProbeConfig setup for testing.  Note that we don't actually install slurm (or other batch systems).  We could install our slurm from devops-itb, although even when i do that, slurm_meter gives an error because /etc/slurm/slurm.conf is not set up.  Not sure how far you'd like to explore this.

brianhlin:
Does `condor_meter` work ok?

test/README.docker.md
```diff
@@ -0,0 +1,17 @@
...
+From the top-level gratia-probe directory, build the image with:
+
+```
+docker build -t gratia-probe-testing -f test/Dockerfile .
+```
+
+And run with:
+
+```
+docker run -it gratia-probe-testing
```
brianhlin, line 15: @edquist what are you imagining as the dev workflow for this? Will a developer rebuild the container each time they make changes then run the script they're testing directly with something like this?

```
docker run -it gratia-probe-testing /usr/share/gratia/condor/condor_meter
```

What do you do if you need to check logs?

brianhlin (approved):

test/README.docker.md
brianhlin: I'd like this `README.md` to be a bit more fleshed out (think about how you would explain how to use this to a student) but it can go in a separate PR.

PR #67, opened 2020-10-19
convert Popen output to str (SOFTWARE-4285)

edquist:


slurm/SlurmProbe.py
```diff
@@ -144,6 +144,8 @@ def get_slurm_version(self):
         cmd = [prog, "--version"]
         p = subprocess.Popen(cmd, stdout=subprocess.PIPE)
         output, _ = p.communicate()
+        if not isinstance(output, str):
+            output = output.decode()
```
matyasselmeci, line 148:
```suggestion
        if not isinstance(output, str):
            output = output.decode(errors="ignore")
```
This will omit bad characters from the variable instead of raising a UnicodeDecodeError. (That being said, what are the chances?)

matyasselmeci (approved):
Looks fine; make my suggested change if you want but it's not necessary.

slurm/SlurmProbe.py
edquist: That is good to know, in general... Though in this case if `srun --version` emits invalid utf-8 bytes (let alone non-ascii bytes at all), arguably that is enough of a surprise to warrant an exception.

Ah, choices.  I guess we can keep it for reference  :)

PR #68, opened 2020-10-21
Convert common to be Python 2/3 compatible (SOFTWARE-4288)

edquist:
FINGERS CROSSED, as they say

### DIFF
--- a/common/DebugPrint
+++ b/common/DebugPrint
@@ -9,6 +9,8 @@ DebugPrint.py
        cat message.txt | DebugPrint.py [-l #|--level=#] [-c <probeconfig>|--conf=<probeconfig>]
 """
 
+from __future__ import print_function
+
 import sys
 from gratia.common import GratiaCore
 import getopt
@@ -18,9 +20,9 @@ def Usage():
     """
     Print the usage.
     """
-    print """usage: DebugPrint.py [-h|--help]
+    print("""usage: DebugPrint.py [-h|--help]
        DebugPrint.py [-l #|--level=#] [-c <probeconfig>|--conf=<probeconfig>] <message>
-       cat message.txt | DebugPrint.py [-l #|--level=#] [-c <probeconfig>|--conf=<probeconfig>]"""
+       cat message.txt | DebugPrint.py [-l #|--level=#] [-c <probeconfig>|--conf=<probeconfig>]""")
 
 def main():
     """
@@ -50,9 +52,9 @@ def main():
     GratiaCore.quiet = 1
 
     if customConfig:
-    	GratiaCore.Initialize(customConfig)
+        GratiaCore.Initialize(customConfig)
     else:
-    	GratiaCore.Initialize()
+        GratiaCore.Initialize()
 
     GratiaCore.quiet = 0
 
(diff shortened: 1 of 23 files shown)


matyasselmeci:
Looks fine to me but apparently doesn't look fine to the tests...

brianhlin:
Does this work with your test Dockerfile?

edquist:
Made a few fixes, uncovered with pylint3.

... To errors i introduced fixing errors from a previous pylint3 run :-p

It's looking cleaner now, except for this one curious part that shows up in the python2 pylint -E checks also.  @matyasselmeci , have you encountered `super()` errors like this before?  The fact that pylint complains about it for python2 also makes me less concerned though.

```
************* Module gratia.common.Gratia
E: 68, 8: Bad first argument 'self.__class__' given to super() (bad-super-call)
E:443, 8: Bad first argument 'self.__class__' given to super() (bad-super-call)
```

@matyasselmeci & @brianhlin: apparently i cannot re-request a review because you have not submitted a review yet - you just made normal (non-review) comments.  I guess github takes that to mean the original review request is still in effect, or something.

matyasselmeci:
Well, I've never tried using `self.__class__` in a `super()` call. There is a potential for harm there since you don't know _for sure_ what the class of `self` is. For example, you can do the following:
```python
bar = Bar()
foo = Foo.__init__(bar)
```
in which case, `self.__class__` would be `Bar` not `Foo` and you'd end up referring to the wrong superclass.

That being said, "if you do it, you deserve it" so I'm not too worried about it in practice. In Python 2, that line would be
```
super(UsageRecord, self).__init__()
```
which works in both Pythons but is repetitive so in Python 3, this works:
```
super().__init__()
```

edquist:
Yeah, i am not convinced that in [this case](https://github.com/opensciencegrid/gratia-probe/blob/0f4e18e7ef07be6dd2fca759b68c75260cc78e62/common/gratia/common/Gratia.py#L68) there is any difference between the existing
```
super(self.__class__, self).__init__()
```
vs
```
super(UsageRecord, self).__init__()
```
vs just
```
record.Record.__init__(self)
```

But it's confusing why pylint/pylint3 call it an _error_.

Anyway i'm inclined to leave it as is.

matyasselmeci:
Yeah I'd call it a warning myself. :shrug:

matyasselmeci (approved):
Looks good to me.

matyasselmeci:
@brianhlin , do you want to take a look as well, or should I merge it?

brianhlin:
Thanks for the review @matyasselmeci! I asked @edquist to run this through his [test Dockerfile](https://github.com/opensciencegrid/gratia-probe/blob/master/test/Dockerfile) to at least verify the `condor_meter` Python 2 functionality before merging. If that goes well, go ahead and merge @edquist!

edquist:
Running condor_meter (with the test history files put in place, under /var/lib/gratia/data) works as well with these python2/3 changes as it does without them.  It prepares the records to send without issue, although both before and after these changes there are "No route to host" errors trying to connect to gratia-osg-itb.opensciencegrid.org.  But i get that from my workstation also, which makes me think it's just an issue with this endpoint.

brianhlin (approved):
LGTM!

PR #71, opened 2020-10-29
convert condor_meter to be python2/3 compatible (SOFTWARE-4283)

edquist:
And, incidentally, introduce utils.bytes2str helper for py2/3 compat, for use in condor_meter, and other already converted spots

...

Other than decoding the output from Popen.communicate(), i don't see anything non python3 in condor_meter

### DIFF
--- a/common/gratia/common/connect_utils.py
+++ b/common/gratia/common/connect_utils.py
@@ -276,9 +276,7 @@ def postRequest(myconnection, to, what, headers):
     myconnection.request('POST', to, what, headers)
     DebugPrint(4, 'DEBUG: POST: OK')
     DebugPrint(4, 'DEBUG: Read response')
-    responseString = myconnection.getresponse().read()
-    if not isinstance(responseString, str):
-        responseString = responseString.decode(errors='ignore')
+    responseString = utils.bytes2str(myconnection.getresponse().read())
     DebugPrint(4, 'DEBUG: Read response: OK')
     
     signal.alarm(0)
--- a/common/gratia/common/probe_config.py
+++ b/common/gratia/common/probe_config.py
@@ -151,9 +151,7 @@ class ProbeConfiguration:
                                        'not really')])
         headers = {'Content-type': 'application/x-www-form-urlencoded'}
         qconnection.request('POST', self.get_RegistrationService(), queryString, headers)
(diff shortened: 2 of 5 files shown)


matyasselmeci:
Sure I suppose. Note that the name "ignore" for the error handling is super misleading. What it really does is omit bad byte sequences from the converted result -- this means it will not survive a round trip. Is this what you want?

edquist:
>Is this what you want?

What i actually want is python2, forever.

Is there actually a way to guarantee unaltered round-trips for bytes->unicode->bytes, for bad utf-8?

Anyway on some level i'm ok with it since i don't think we really expect non-ascii bytes, certainly not bad utf-8 bytes.  (In this case from `condor_config_val`.)

I'm open to being told otherwise though.

matyasselmeci:
Well, if you decode as latin-1 and encode as latin-1 and don't do any text transformations like .lower() or .upper() on the str, then you can survive the round trip.

Here's how all the different methods handle bad unicode:
```
>>> badunicode = b"A-\x80\x81-Z"
>>> badunicode.decode(errors="strict")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 2: invalid start byte
>>> badunicode.decode(errors="replace")
'A-��-Z'
>>> badunicode.decode(errors="ignore")
'A--Z'
>>> badunicode.decode("latin-1")
'A-\x80\x81-Z'
>>> badunicode.decode(errors="surrogateescape")
'A-\udc80\udc81-Z'
```
("surrogateescape" is Python 3 only but that can also give you a round-trip)

matyasselmeci (approved):

matyasselmeci:
Take your pick or feel free to merge as-is.

edquist:
For now i think i'm going to merge as-is.

In principle, i would prefer either to explicitly expect the input to be in utf-8 (or just ascii), or else to treat it as binary, in which case latin-1 would be appropriate.  But in the latter case we'd then also have to ensure that the output file handles everywhere use latin-1 encoding (otherwise they'd default to be interpreted as utf-8), and that's not something i feel great about trying to chase down at the moment.

PR #72, opened 2020-11-05
Explicitly define __python (3) for el8 (SOFTWARE-4348)

edquist:


matyasselmeci:
You'll have to do one other thing before RPM builds for EL8 pass: you'll have to change the shebang lines. I've been using this code for that in the %install section:
```
%if 0%{?rhel} >= 8
find . -type f -exec sed -ri '1s,^#!\s*(/usr)?/bin/(env *)?python.*,#!%{__python},' '{}' +
%endif
```
feel free to critique my regex or modify as needed.

edquist:
Ah, good catch!  Thank you!

edquist:
Just to get my thoughts down on paper here ...

> ```
> %if 0%{?rhel} >= 8
> find . -type f -exec sed -ri '1s,^#!\s*(/usr)?/bin/(env *)?python.*,#!%{__python},' '{}' +
> %endif
> ```

I would be more hesitant about the `find .` part if it were run in a git repo (affecting files under the `.git` dir), but in the `%install` section i believe it's just the archive contents.

I also am a bit inclined to update (with sed) only the files that have the matching line in question; something like

```
find . -type f -exec egrep -lZ '^#!\s*(/usr)?/bin/(env *)?python\>' |
xargs -0 sed -ri '1s,^#!\s*(/usr)?/bin/(env *)?python.*,#!%{__python},'
```

otherwise sed rewrites _all_ files, even the ones that don't have the substitution applied.  But i guess it's only a small inefficiency (in a relatively small repo, after all).  Though it does mean all files will have their mtime updated, even the non-python files.

Anyway, i'm fine with using what you've got for consistency.

matyasselmeci:
Is there a way to make grep stop at line 1 if there's no match? Looks like otherwise it reads the whole file (which, to be sure, is a lesser inefficiency than writing the whole file).

matyasselmeci (approved):
Looks good to me. Apparently you broke GitHub's syntax highlighting though...

PR #74, opened 2020-11-10
Fix condor-python version for el8 (SOFTWARE-4348)

edquist:
And, from my discussion with @brianhlin this morning, apparently we will go python3 for el7, also.


matyasselmeci (approved):
LGTM.

PR #75, opened 2020-11-11
add python2/3 compat for common2 (SOFTWARE-4348)

edquist:
the el7 build started breaking when i made python3 explicit for el7, due to the common2 stuff.  So i guess we'll try to convert it also.

@matyasselmeci, do you know if the `from .` things that 2to3 suggested will work for python2 also?


matyasselmeci:
In Python 2 you have to enable the `from .` syntax by doing `from __future__ import absolute_import` at the top.

edquist:
thanks @matyasselmeci - i've added `from __future__ import absolute_import` lines.

Yes there do not appear to be any system libs with the same names as the .py modules here.

common2/gratia/common2/pgpinput.py
```diff
@@ -10,10 +12,10 @@
     import uuid  # for unique cursor name
 except ImportError:
     # for python < 2.5 (uuid not available)
-    import uuid_replacement as uuid
+    from . import uuid_replacement as uuid
```
matyasselmeci, line 15: If the comment above is accurate, then you can replace the whole try/except block with
```
import uuid
```

common2/gratia/common2/timeutil.py
```diff
@@ -26,7 +28,7 @@
     from gratia.common.debug import DebugPrint
 except ImportError:
     def DebugPrint(val, msg):
-        print ("DEBUG LEVEL %s: %s" % (val, msg))
+        print(("DEBUG LEVEL %s: %s" % (val, msg)))
```
matyasselmeci, line 31:
```suggestion
        print("DEBUG LEVEL %s: %s" % (val, msg))
```
Looks like the converter didn't notice that it was already using parens...

common2/gratia/common2/timeutil.py
```diff
@@ -87,24 +89,24 @@ def strfdelta(tsec, format_str="P", format_no_day="PT{H}H{M}M{S}S", format_zero=
...
-                if i in l.keys():
+                if i in list(l.keys()):
                     d[i], rem = divmod(rem, l[i])
                     if d[i] != 0:
                         format_str = "%s{%s}%s" % (format_str, i, i)
         else:
             if 0 < tsec < 86400:
                 format_str = format_no_day
-            k = map(lambda x: x[1], list(f.parse(format_str)))
+            k = [x[1] for x in list(f.parse(format_str))]
```
matyasselmeci, line 106:
```suggestion
            k = [x[1] for x in f.parse(format_str)]
```
Since you're iterating over it, you don't need to convert it to a list first.

matyasselmeci (approved):
There's a few fixups but overall good so pre-approving.

common2/gratia/common2/pgpinput.py
edquist: Oh goodness you are right.

common2/gratia/common2/timeutil.py
edquist: That is true, but i left it in because that stylistic oddity existed in the original  :)

Nevertheless i can't otherwise justify keeping it.

PR #78, opened 2020-11-30
fix stderr handling in communicate() (SOFTWARE-4348)

edquist:
since stderr is not captured, communicate returns None for stderr,
which is not handled by bytes2str.  The stderr variables here are unused,
so just drop them.


brianhlin (approved):
LGTM! Please hotpatch ASU's CE to see if this fixes it and we get the records that we're missing

edquist:
Seems to fix it on the ASU CE

PR #79, opened 2020-12-04
Add multisite support for pilot container probe (SOFTWARE-4386)

edquist:
- set ProbeName based on SiteName
- bucket records based on ProbeName/SiteName
- reconfigure probe per site in child processes before sending

Not yet tested

### DIFF
--- a/osg-pilot-container/osgpilot_meter
+++ b/osg-pilot-container/osgpilot_meter
@@ -3,6 +3,7 @@
 from __future__ import print_function
 
 import os
+import re
 import time
 import random
 import signal
@@ -458,8 +459,8 @@ def query_and_update_db(opts):
 
 
 def site_ProbeName(site):
-    sitehash = sha1sum(site)[:7]
-    return "osg-pilot-container:%s.gratia.opensciencegrid.org" % sitehash
+    site_dns = re.sub(r'[^a-zA-Z0-9-]', '-', site).strip('-')  # sanitize site
+    return "osg-pilot-container:%s.gratia.opensciencegrid.org" % site_dns
 
 
 def outbox_row_to_jur(row):
--- a/osg-pilot-container/osgpilot_meter
+++ b/osg-pilot-container/osgpilot_meter
@@ -5,6 +5,7 @@ from __future__ import print_function
 import os
 import time
 import random
+import signal
 import hashlib
 import sqlite3
 import optparse
--- a/osg-pilot-container/osgpilot_meter
+++ b/osg-pilot-container/osgpilot_meter
@@ -660,6 +660,8 @@ def send_alternate_records_child(info, record_list, gratia_config):
     DebugPrint(2, "Number of usage records submitted: %d" % count_submit)
     DebugPrint(2, "Number of usage records found: %d" % count_found)
 
+    GratiaCore.ProcessCurrentBundle()
+
     os._exit(0)
 
 
--- a/osg-pilot-container/osgpilot_meter
+++ b/osg-pilot-container/osgpilot_meter
@@ -603,10 +603,7 @@ def send_alternate_records(gratia_info, gratia_config):
                 os._exit(0)
             os._exit(0)
         else: # I am parent
-            try:
-                os.waitpid(pid, 0)
-            except:
-                raise
+            os.waitpid(pid, 0)
 
 
 def send_alternate_records_child(info, record_list, gratia_config):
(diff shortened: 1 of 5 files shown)


edquist:
Apparently there were a number of python3 compatibility issues, discovered in testing, which i seem to have fixed.  Apparently this means the version of this probe running in tiger previously was before the python3 changes for gratia-probe were merged in.  Anyway this is good that it came up (in order to fix it), because there were a couple places that were probably broken for other probes as well.

Things seemed to be working in testing, although i have not seen any records show up in gracc yet - possibly just a matter of pilot jobs finishing...

edquist:
Multi-site functionality is working now in the tiger instance

https://gracc.opensciencegrid.org/kibana/goto/51058d1b867c5d59d1551e56e41e03c4

Would like to merge this in and do the daily partial record work for SOFTWARE-4404 in a separate branch.

osg-pilot-container/osgpilot_meter
```diff
@@ -579,6 +584,75 @@ def main():
...
+            except Exception as e:
+                DebugPrint(2, "Failed to send alternate records: %s" % str(e))
+                DebugPrintTraceback(2)
+                os._exit(0)
+            os._exit(0)
+        else: # I am parent
+            try:
+                os.waitpid(pid, 0)
+            except:
+                raise
```
brianhlin, line 609: Why does this need to be in a try/except if we're just raising?

brianhlin (approved):
Sure, looks generally fine to me!

djw8605:
The gracc link you sent has a nonsense ProbeName.  Should it include the SiteName somehow?

osg-pilot-container/osgpilot_meter
edquist: Yeah it looks like the try/except there does not accomplish anything.

So, I initially lifted the alternate-site code directly from condor_meter - I tried only to adjust the parts that actually needed to be changed for this probe's implementation rather than trying to write a leaner replacement.  Apparently Edgar did the original multi-site support work in the condor probe for GRATIA-188 in 79cb619 ([here](https://github.com/opensciencegrid/gratia-probe/commit/79cb6197ee1bc40fccaf7c73c7e05ba5dd7770fb#diff-c96466053a3ee99c9d454bf9db438662de6f32ab1f82e356038345038815ad8bR929-R992)).

I am not opposed to cleaning it up though.

edquist:
> The gracc link you sent has a nonsense ProbeName. Should it include the SiteName somehow?

Hm.  For me it's just a search for `ProbeName: osg-pilot-container\:*`.  And if you break that up by SiteName then you see JetStream, Purdue Geddes, ISI, and CHTC

djw8605:
The ProbeName is "osg-pilot-container:1974f7b.gratia.opensciencegrid.org", which is nonsense, right?

edquist:
> The ProbeName is "osg-pilot-container:1974f7b.gratia.opensciencegrid.org", which is nonsense, right?

Oh i see what you mean now - no, the nonsense-looking part is a hash of the SiteName.

For [SOFTWARE-4386](https://opensciencegrid.atlassian.net/browse/SOFTWARE-4386) BrianB had suggested we do a pattern like that where the ProbeName is based on the SiteName by containing a hash of the SiteName.  The ProbeName is computed for a given SiteName [here](https://github.com/opensciencegrid/gratia-probe/pull/79/files#diff-f5aee1b71bdecdb74c1a431acc1770cc2171e74573ef0ae294f57732fce5a10bR460-R462).

edquist:
If it would make things more obvious, we could also include a sanitized version of the SiteName in the ProbeName, eg after the hash.

djw8605:
I feel like a hash of the probe name is needlessly obfuscating the ProbeName.  Just sanitize the SiteName to make it dns-like name, and put it in the ProbeName.  If the ProbeName is always going to be a hash, then there is no advantage to at least making the probename a little readable.

edquist:
(And by "sanitized" i mean something like `re.sub(r'[^\w]', '_', SiteName).lower()`)

djw8605:
Yeah, something like that.  Something approaching urlencode, but without the %... stuff.

edquist:
Per discussion with @djw8605, i've replaced the SiteName hash from the ProbeName calculation with a sanitized version of the SiteName, converting non-alphanumerics to dashes, to make it appropriate for a dns name/fqdn, but still keep the site be easily identifiable for things that look at probe name.

So for instance now the mapping does:

```
>>> site_ProbeName("Purdue Geddes")
'osg-pilot-container:Purdue-Geddes.gratia.opensciencegrid.org'
```

PR #81, opened 2020-12-16
fix python3 packaging deps for el7 vs el8 (SOFTWARE-4398)

edquist:
Derp, discovered in testing[1], the python openssl package name changes between el7/el8, and there do not appear to be any common virtual capabilities provided...

[1] https://osg-sw-submit.chtc.wisc.edu/tests/20201216-1241/packages.html


edquist:
Same for `python36-mysql` and `python36-pytz`

matyasselmeci (approved):

PR #83, opened 2020-12-23
Send daily partial records (SOFTWARE-4404)

edquist:
The concept here is:

- If a job has been active for > 1 day:
  - send it as if it were completed
  - reset its start time, total cpu, and max memory aggregates

Note that this gives rolling updates throughout the day as partial jobs cross their 24-hour lifetimes.  This is not quite the same thing as having a separate "daily flush" code path, where all unfinished jobs would get sent and reset as if they were finished, regardless of their lifetime.  But the main point is the same, that the progress for each job will be reported daily.

I think the code is likely complete at this point, but given that it's entirely untested i am marking this PR as a draft.

### DIFF
--- a/osg-pilot-container/osgpilot_meter
+++ b/osg-pilot-container/osgpilot_meter
@@ -36,9 +36,10 @@ LOCAL_USER = "osgvo-container-pilot"
 
 grid_type = "OSG"
 
-DONE_IF_NOT_SEEN_AFTER = 1 * 60 * 60;
-
-POLL_INTERVAL = 15 * 60  # seconds
+# in seconds
+DONE_IF_NOT_SEEN_AFTER =  1 * 60 * 60
+POLL_INTERVAL          = 15 * 60
+PARTIAL_RECORD_CUTOFF  = 24 * 60 * 60
 
 # log levels
 CRITICAL = 0
@@ -375,6 +376,10 @@ def detect_finished_jobs(sqldb, current_ts):
 
     DebugPrint(INFO, "- inserted %s rows into outbox" % rc1)
     DebugPrint(INFO, "- deleted %s rows from jobs" % rc2)
+
+
+def report_partial_jobs(sqldb):
+    pass
 
 
 def update_active_jobs(sqldb):
--- a/osg-pilot-container/osgpilot_meter
+++ b/osg-pilot-container/osgpilot_meter
@@ -477,8 +477,8 @@ def detect_and_update_partial_jobs(sqldb):
          where last_updated - DaemonStartTime >= {partial_record_cutoff};
     """.format(partial_record_cutoff=PARTIAL_RECORD_CUTOFF)
 
-    rc1 = sqldb.execute(insert_sql, (dead_cutoff,)).rowcount
-    rc2 = sqldb.execute(update_sql, (dead_cutoff,)).rowcount
+    rc1 = sqldb.execute(insert_sql).rowcount
+    rc2 = sqldb.execute(update_sql).rowcount
 
     DebugPrint(INFO, "- inserted %s rows into outbox" % rc1)
     DebugPrint(INFO, "- updated %s rows from jobs" % rc2)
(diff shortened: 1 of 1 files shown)


osg-pilot-container/osgpilot_meter
```diff
@@ -440,6 +459,30 @@ def detect_new_jobs(sqldb, current_ts):
...
+    # reset aggregates
+    update_sql = """
+        update jobs
+           set DaemonStartTime = last_updated,
+               total_cpu = 0,
+               MemoryUsage = 0
+         where last_updated - DaemonStartTime >= {partial_record_cutoff};
+    """.format(partial_record_cutoff=PARTIAL_RECORD_CUTOFF)
+
+    rc1 = sqldb.execute(insert_sql, (dead_cutoff,)).rowcount
```
bbockelm, line 480: I know that we most likely are only running a single instance at a time but can you humor the computer scientist in me and do this statement and the next one in a single transaction?

osg-pilot-container/osgpilot_meter
```diff
@@ -440,6 +459,30 @@ def detect_new_jobs(sqldb, current_ts):
...
+    rc2 = sqldb.execute(update_sql, (dead_cutoff,)).rowcount
+
+    DebugPrint(INFO, "- inserted %s rows into outbox" % rc1)
```
bbockelm, line 483: Do we make the assumption that the row count for these two queries are always the same?  If so, we should log an error if they are different.

osg-pilot-container/osgpilot_meter
```diff
@@ -440,6 +459,30 @@ def detect_new_jobs(sqldb, current_ts):
...
+    update_sql = """
+        update jobs
+           set DaemonStartTime = last_updated,
+               total_cpu = 0,
+               MemoryUsage = 0
+         where last_updated - DaemonStartTime >= {partial_record_cutoff};
+    """.format(partial_record_cutoff=PARTIAL_RECORD_CUTOFF)
+
+    rc1 = sqldb.execute(insert_sql, (dead_cutoff,)).rowcount
+    rc2 = sqldb.execute(update_sql, (dead_cutoff,)).rowcount
```
bbockelm, line 481: What happens if the job makes the cutoff for one query but not the other (suppose that the OS sleeps the process for 2 hours between these two SQL statements)?  Do we just loose the hours?

osg-pilot-container/osgpilot_meter
edquist: This is done in a single transaction; the commit comes later ([here](https://github.com/opensciencegrid/gratia-probe/pull/83/files#diff-f5aee1b71bdecdb74c1a431acc1770cc2171e74573ef0ae294f57732fce5a10bR502)).

osg-pilot-container/osgpilot_meter
edquist: The timestamp for all incoming records is the same in the inbox table, then aggregated into the jobs table.  Both queries use the same `last_updated - DaemonStartTime >= {partial_record_cutoff}` condition from the jobs table, not the current query execution timestamp.  So, adding a 2 hour sleep between those two execute statements (or anywhere else, for that matter) should not affect the results.

osg-pilot-container/osgpilot_meter
edquist: Eh, i guess there is no logic that depends on these two row counts being the same.  But the concept is that the same rows are both copied and then updated, so they ought to match.  They do use the same where clause on the same table while in a transaction.  If it makes you more comfortable to see an error in the case that these don't match, the sure i can add a check for that.

edquist:
Testing looked good today, after one fix.  All jobs over a day old got reported as completed at the probe run time, and were reset with their start times set to that time.

osg-pilot-container/osgpilot_meter
edquist: Added a row count check with error log message per Brian's suggestion

edquist (commented):
Just to note it here, BrianB gave a verbal approval this morning.

PR #84, opened 2021-01-07
fix python3 integer division for duration formatter

edquist:
value, minutes, and hours here are expected to be ints, but '/' does
float division in python3


djw8605 (approved):
Well that's a simple change.

PR #87, opened 2021-02-08
Fix typo in match expression attr name

brianhlin:
As noticed by Mat


edquist (approved):
`MATCH_EXP_JOBGLIDEIN_ResourceName` matches what i see in the actual condor_meter code

PR #89, opened 2021-02-09
add 'GratiaVersion' to UsageRecords (SOFTWARE-4455)

edquist:
Where GratiaVersion is the rpm version-release followed by the short hash of the git commit.

I did not actually make GratiaVersion a function, since i don't intend that we'd ever call `r.GratiaVersion(...)` explicitly.

Also kind of tidy up the `%%%RPMVERSION%%%` shell find & replace.  (According to me, maybe.)

Scratch builds succeeded: https://koji.opensciencegrid.org/koji/taskinfo?taskID=328312

and in Gratia.py from the built rpm i see

```
    __GratiaVersion = "1.22.3-1.osg35.el8.0cf7ca1"
```

### DIFF
--- a/rpm/gratia-probe.spec
+++ b/rpm/gratia-probe.spec
@@ -14,7 +14,11 @@ BuildRequires:      gcc-c++
 BuildRequires:      python3
 %endif
 
+%if 0%{?rhel} < 8
 BuildRequires:      git
+%else
+BuildRequires:      git-core
+%endif
 
 # just do a single arch build until we drop the compiled tool in pbs-lsf
 ExcludeArch: noarch
--- a/rpm/gratia-probe.spec
+++ b/rpm/gratia-probe.spec
@@ -339,7 +339,7 @@ git_commit_id=$(gzip -d < %{SOURCE0} | git get-tar-commit-id)
 rpmver=%{version}-%{release}.${git_commit_id:0:7}
 find $RPM_BUILD_ROOT%{_datadir}/gratia $RPM_BUILD_ROOT%{python_sitelib} \
   -type f -exec fgrep -ZIle '%%%%%%RPMVERSION%%%%%%' {} + | xargs -0 \
-  perl -wpi -e "s&%%%%%%RPMVERSION%%%%%%&$rpmver&g"
+  sed -i "s&%%%%%%RPMVERSION%%%%%%&$rpmver&g"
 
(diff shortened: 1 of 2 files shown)


djw8605:
Any chance we can remove the calling of perl?  One less thing that depends on perl.

edquist:
FWIW i did not _add_ this perl dependency.

How ironic that i, of all people, would remove it.

I don't know what you people have against Perl.

But yeah, sed works just fine here instead.

djw8605 (approved):
With the removal of perl, I'm happy.

rpm/gratia-probe.spec
```diff
@@ -330,10 +336,10 @@ install -d $RPM_BUILD_ROOT/%{_sysconfdir}/gratia
 
 
 # Burn in the RPM version into the python files.
-grep -rIle '%%%%%%RPMVERSION%%%%%%' $RPM_BUILD_ROOT%{_datadir}/gratia $RPM_BUILD_ROOT%{python_sitelib} | while read file; do \
-  perl -wpi.orig -e 's&%%%%%%RPMVERSION%%%%%%&%{version}-%{release}&g' "$file" && \
-    rm -fv "$file.orig"
-done
+rpmver=%{version}-%{release}.${git_commit_id:0:7}
+find $RPM_BUILD_ROOT%{_datadir}/gratia $RPM_BUILD_ROOT%{python_sitelib} \
+  -type f -exec fgrep -ZIle '%%%%%%RPMVERSION%%%%%%' {} + | xargs -0 \
```
matyasselmeci, line 341: I'm afraid to ask why you need 6 `%` on each side here but 3 `%` on each side in Gratia.py

matyasselmeci: Oh right, specfile macros. never mind.

rpm/gratia-probe.spec
```diff
@@ -14,6 +14,8 @@ BuildRequires:      gcc-c++
 BuildRequires:      python3
 %endif
 
+BuildRequires:      git
```
matyasselmeci, line 17:
```suggestion
%if 0%{?rhel} < 8
BuildRequires:      git
%else
BuildRequires:      git-core
%endif
```
Gives you a somewhat smaller buildroot (which means a faster build).

matyasselmeci (approved):
Looks good; one optional suggestion.

PR #91, opened 2021-02-10
elevate lots of probes to museum status (SOFTWARE-4467)

edquist:


edquist:
Note scratch builds look good

https://koji.opensciencegrid.org/koji/taskinfo?taskID=329827
https://koji.opensciencegrid.org/koji/taskinfo?taskID=329828

djw8605:
Some comments:
- For SGE and LSF, I don't think we have any method of gathering accounting information.  We can gather rough wallclock time, but nothing else.
- Check with FNAL if they are using the dCache probe.  We still get transfer records from their dCache installation.  Doesn't mean we need to keep it, but good to notify stakeholders

edquist:
Thanks @djw8605  for noting these points.

@brianhlin not sure if you have any thoughts about Derek's items.

I'll also note (in case it softens the blow) that this PR targets the `opensciencegrid:2.x` branch intended for OSG 3.6 - separate from `opensciencegrid:master`, which will continue to serve gratia-probe 1.x releases for OSG 3.5.

brianhlin:
@djw8605 

> For SGE and LSF, I don't think we have any method of gathering accounting information. We can gather rough wallclock time, but nothing else.

I assume that this is because the blahp isn't dumping this info back into the HTCondor job ad?

> Check with FNAL if they are using the dCache probe. We still get transfer records from their dCache installation. Doesn't mean we need to keep it, but good to notify stakeholders

Any idea who's sending these from FNAL?

djw8605:
Right, the blahp isn't dumping the info back.  Now that I think about it, I don't think PBS is supported either.  I think I only wrote the SLURM one using `sacct`.

I don't know who is sending the data from FNAL.  Someone in their storage group is all I know.

brianhlin:
> Right, the blahp isn't dumping the info back. Now that I think about it, I don't think PBS is supported either. I think I only wrote the SLURM one using sacct.

@edquist took a look at HTCondor-CE vs PBS probe records for a PBS backend (https://opensciencegrid.atlassian.net/browse/SOFTWARE-4444?focusedCommentId=361971) and we were only missing some info that we decided we were ok losing for the time being. Are you saying that some of the attributes in the HTCondor-CE case are off?

retzkek:
Yes, our storage group still runs the dCache and enstore probes, reporting to our local gracc instance. We then forward transfers to OSG. Does OSG no longer want to see these transfers?

List of non-condor probes still used at FNAL:

onevm48
enstore-transfer
enstore-tapedrive
enstore-storage
dCache-transfer
dCache-storagegroup

Do these probes really need to be deleted?

edquist:
What do you need, besides a miracle?

Probes.  Lots of probes.

brianhlin:
> We then forward transfers to OSG. Does OSG no longer want to see these transfers?

@djw8605 can correct me if I'm wrong but I don't think we use these records at all. Does the storage group need them for internal accounting?

> Do these probes really need to be deleted?

Turning this one around, why are these probes still needed?

retzkek:
>  I don't think we use these records at all.

There used to be graphs and stats about transfers floating around (which I always groaned about since FNAL and BNL were the only institutions reporting transfers), but maybe nobody looks at that any more?

>  Does the storage group need them for internal accounting? 

Yes, the dCache and enstore data is primarily used for internal accounting, for example: tracking how many tapes OSG GRACC data is using.

The onevm probe is used for fermicloud accounting. I see a handful of VMs being run by OSG folks, so I think it's fair to say OSG has a stake in this as well.

Some of these probes will eventually be made redundant or no longer needed, e.g. fermicloud is moving away from opennebula. We can make plans to move any other functionality or fork the probes, but it will take some time.

osg-cat:
Can we avoid removing storage/transfer probes until Frank has had the chance to weigh in? He is really the one who leads the OSDF effort and cares the most about usable statistics about it.

brianhlin:
@retzkek I'm happy to keep them around! I just wanted to be aggressive about removal as a starting point. Do you also need `dCache-storage`?

@edquist please reinstate the probes that Kevin noted above:

> onevm48
enstore-transfer
enstore-tapedrive
enstore-storage
dCache-transfer
dCache-storagegroup

edquist:
>please reinstate the probes that Kevin noted above

Misc Probes: Your membership to The Continental Museum has been ... Revoked

retzkek:
>  I'm happy to keep them around! I just wanted to be aggressive about removal as a starting point. Do you also need dCache-storage?

Thanks. I don't see any records from a dCache-storage probe, so if we ever used it nobody noticed it stopped, so I'll say no, we don't.

edquist:
For now I've added back

- dCache-storagegroup
- dCache-transfer
- enstore-storage
- enstore-tapedrive
- enstore-transfer
- onevm

notably leaving out dcache-storage.  We will wait back to hear what Frank wants for the other storage/trasnfer probes.

edquist:
> We will wait back to hear what Frank wants for the other storage/trasnfer probes.

Frank wanted us to keep the transfer probes, so gridftp/xrootd-transfer were added back.

Had to manually resolve the merge for the htcondor-ce probe customizations.

After some random fixes, scratch builds are looking good again:

https://koji.opensciencegrid.org/koji/taskinfo?taskID=331910
https://koji.opensciencegrid.org/koji/taskinfo?taskID=331909

@brianhlin @djw8605 lmk if you think anything else is needed here.

edquist:
@brianhlin says no gridftp-transfer probe, so i've removed them.

New scratch builds:

https://koji.opensciencegrid.org/koji/taskinfo?taskID=331950
https://koji.opensciencegrid.org/koji/taskinfo?taskID=331948

edquist:
I got a verbal :+1: from @brianhlin and that is good enough for me

PR #92, opened 2021-02-18
fix python3 int division in niceNum

edquist:
@djw8605 discovered unhandled exceptions which traced back to a `/` in niceNum, but which still relied on the python2 int division behavior


djw8605 (approved):
LGTM

PR #94, opened 2021-02-24
Fix searching for current attribute when adding to the attribute list

djw8605:
The verbatim append to list above always adds a space, so include that in the search
otherwise setting "Queue" after "QueueTime" would cause "QueueTime" to be removed from the list


edquist:
Good catch.  Had to stare at the filter expression for a while to understand the logic but it looks right.

edquist (approved):

PR #96, opened 2021-05-13
use AuthTokenSubject attr for SciToken VO info (SOFTWARE-4615)

edquist:
Is it this easy?

And, is it OK to use the same `AuthTokenSubject` attr for both VOName and ReportableVOName ?

### DIFF
--- a/common/gratia/common/condor_ce.py
+++ b/common/gratia/common/condor_ce.py
@@ -375,7 +375,4 @@ def queryJob(jobid):
         certinfo['DN'] = info['AuthTokenSubject']
     if 'x509UserProxyFirstFQAN' in info:
         certinfo['FQAN'] = info['x509UserProxyFirstFQAN']
-    elif 'AuthTokenIssuer' in info:
-        certinfo["FQAN"] = info['AuthTokenIssuer']
     return certinfo
-
--- a/common/gratia/common/condor_ce.py
+++ b/common/gratia/common/condor_ce.py
@@ -375,6 +375,9 @@ def queryJob(jobid):
         certinfo['DN'] = info['AuthTokenSubject']
     if 'x509UserProxyFirstFQAN' in info:
         certinfo['FQAN'] = info['x509UserProxyFirstFQAN']
-    # intentionally omitting AuthTokenIssuer equivalent for FQAN; see PR #96
+    # FQAN is used as a backup for DN/VO in the certinfo to populate
+    # VOName/ReportableVOName in the usage record, respectively.
+    # Since DN/VO are always populated in the AuthToken case, we are
+    # intentionally omitting AuthTokenIssuer equivalent for FQAN.
+    # See https://github.com/opensciencegrid/gratia-probe/pull/96#discussion_r651148685
     return certinfo
-
--- a/common/gratia/common/condor_ce.py
+++ b/common/gratia/common/condor_ce.py
@@ -375,4 +375,6 @@ def queryJob(jobid):
         certinfo['DN'] = info['AuthTokenSubject']
     if 'x509UserProxyFirstFQAN' in info:
         certinfo['FQAN'] = info['x509UserProxyFirstFQAN']
+    # intentionally omitting AuthTokenIssuer equivalent for FQAN; see PR #96
     return certinfo
+
--- a/common/gratia/common/condor_ce.py
+++ b/common/gratia/common/condor_ce.py
@@ -142,19 +142,22 @@ def createCertinfoXML(classad):
     uniqID.appendChild(dom.createTextNode(unique_id))
 
     # Sample: x509UserProxyVOName = "cms"
-    voStr = classad.get("x509UserProxyVOName", "")
+    voStr = classad.get("x509UserProxyVOName", "") or \
+            classad.get("AuthTokenIssuer", "")
     vo = dom.createElement("VO")
     vo.appendChild(dom.createTextNode(voStr))
 
     # Sample: x509UserProxyFirstFQAN = "/cms/Role=NULL/Capability=NULL"
-    fqanStr = classad.get("x509UserProxyFirstFQAN", "")
+    fqanStr = classad.get("x509UserProxyFirstFQAN", "") or \
+              classad.get("AuthTokenIssuer", "")
     fqan = dom.createElement("FQAN")
     fqan.appendChild(dom.createTextNode(fqanStr))
 
     # Sample: x509userproxysubject = "/C=TW/O=AP/OU=GRID/CN=Nitish Dhingra 142746"
-    x509Str = classad.get("x509userproxysubject", "")
-    x509 = dom.createElement("DN")
-    x509.appendChild(dom.createTextNode(x509Str))
+    dnStr = classad.get("x509userproxysubject", "") or \
+            classad.get("AuthTokenSubject", "")
+    dn = dom.createElement("DN")
+    dn.appendChild(dom.createTextNode(dnStr))
 
     info = dom.createElement("GratiaCertInfo")
     info.appendChild(dom.createTextNode("\n  "))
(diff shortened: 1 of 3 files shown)


djw8605:
Interesting.  In python, the `or` means it will execute the first statement, if that returns True, not execute the second.  Right?

And, we have the convention that the `sub` in the token is **ALWAYS** the VO?  According the spec, it's suppose to be something linking to the specific user, but I suppose it doesn't have to be.

https://scitokens.org/technical_docs/Claims
> sub (Subject): Typically indicates the individual or entity this token was originally issued to. The subject is considered to be locally unique in the context of the VO. Usage of this claim is OPTIONAL. The token validators SHOULD NOT utilize the subject for authorization or identity mapping. Suggested use cases are auditing, monitoring, or tracing. Due to privacy concerns, some VOs may issue non-human-readable subjects or multiple anonymized subjects per individual. A VO MUST NOT use the same subject for multiple individuals.

edquist:
Hey @djw8605,

>Interesting. In python, the `or` means it will execute the first statement, if that returns True, not execute the second. Right?

Right, that's the idea.  The `and` and `or` from python are short-circuit operators like `&&` and `||` in C and the shell.

So it will use the x509 attributes if present for DN, VOName, and ReportableVOName; but if those are not present it will use the AuthToken attributes.

If you feel like this reads awkwardly I can re-write it - I did it this way because it seemed fairly straightforward, but I'm open to suggestions.

djw8605:
No, I think I'm ok with the code.  I just want to make sure we are on the same page for the handling of `sub` in the token.

edquist:
> And, we have the convention that the `sub` in the token is **ALWAYS** the VO?
> According the spec, it's suppose to be something linking to the specific user, but I suppose it doesn't have to be.

My impression from @bbockelm's comments were that we'd send the subject from the auth token as the VO, but that GRACC would do some sort of processing to turn that into the real VO.  (Brian, did i understand that right?)

bbockelm:
Hi -- not quite.

The subject is the closest equivalent to the DN; the issuer is the closest equivalent to the VO name.  We should map accordingly.

Brian

condor/condor_meter
```diff
@@ -592,9 +594,12 @@ def classadToJUR(classad):
 
     setIfExists(r.LocalUserId, classad, "Owner")
     setIfExists(r.GlobalUsername, classad, "User")
-    setIfExists(r.DN, classad, 'x509userproxysubject')
-    setIfExists(r.VOName, classad, 'x509UserProxyFirstFQAN')
-    setIfExists(r.ReportableVOName, classad, 'x509UserProxyVOName')
+    setIfExists(r.DN, classad, 'x509userproxysubject') or \
+    setIfExists(r.ReportableVOName, classad, 'AuthTokenSubject')
```
brianhlin, line 598:
```suggestion
    setIfExists(r.DN, classad, 'AuthTokenSubject')
```

Copy-paste mistake?

edquist: Good catch, thanks!

brianhlin:
Don't we need corresponding certinfo changes for our non-HTCondor batch systems?

common/gratia/common/condor_ce.py
```diff
@@ -142,19 +142,22 @@ def createCertinfoXML(classad):
...
     fqan.appendChild(dom.createTextNode(fqanStr))
 
     # Sample: x509userproxysubject = "/C=TW/O=AP/OU=GRID/CN=Nitish Dhingra 142746"
-    x509Str = classad.get("x509userproxysubject", "")
-    x509 = dom.createElement("DN")
-    x509.appendChild(dom.createTextNode(x509Str))
+    dnStr = classad.get("x509userproxysubject", "") or \
+            classad.get("AuthTokenSubject", "")
+    dn = dom.createElement("DN")
+    dn.appendChild(dom.createTextNode(dnStr))
```
brianhlin, line 160: Won't all the `classad.get(...) or` expressions always result in the `x509*` attr value or the empty string?

common/gratia/common/condor_ce.py
```diff
@@ -362,9 +367,15 @@ def queryJob(jobid):
...
+    elif 'AuthTokenIssuer' in info:
+        certinfo["VO"] = info['AuthTokenIssuer']
     if 'x509userproxysubject' in info:
         certinfo['DN'] = info['x509userproxysubject']
+    elif 'AuthTokenSubject' in info:
+        certinfo['DN'] = info['AuthTokenSubject']
     if 'x509UserProxyFirstFQAN' in info:
         certinfo['FQAN'] = info['x509UserProxyFirstFQAN']
+    elif 'AuthTokenIssuer' in info:
+        certinfo["FQAN"] = info['AuthTokenIssuer']
```
brianhlin, line 379: I'm not sure this actually makes sense. @djw8605 does the GRACC depend on the certinfo FQAN, which currently corresponds to the first VOMS attr of the proxy

brianhlin (changes requested):

common/gratia/common/condor_ce.py
djw8605: Is this used in determining the VOName or ReportableVOName?

common/gratia/common/condor_ce.py
edquist: Yes.  It defaults to the x509 value, if present, otherwise the empty string evaluates as False in a boolean context, so the short-circuit `or` will fall back to the AuthToken\* value

brianhlin: Oh right

common/gratia/common/condor_ce.py
edquist: So as it stands here, the AuthToken attrs are used as follows -

For certinfo:
- AuthTokenIssuer -> VO & FQAN
- AuthTokenSubject -> DN

For gratia usage records:
- AuthTokenIssuer -> VOName & ReportableVOName
- AuthTokenSubject -> DN

I fuzzily remember that this had been the result of some conversation, but I am having trouble tracking down where the decision about what to do for FQAN happened (if it actually did).  I'm open to input in any case.

brianhlin: @edquist what is the certinfo FQAN used for?

edquist: Ah, @brianhlin, always with the good questions!  :)

https://github.com/opensciencegrid/gratia-probe/blob/v2.0.1-2/common/gratia/common/certinfo.py#L105-L110

So, if there is no DN in the certinfo, then FQAN gets used for VOName (and VO from certinfo gets used for ReportableVOName)

Although, since we put AuthTokenSubject into the certinfo DN, that above path is not taken.  A little bit further down you see the same result though (that is, if DN _is_ in certinfo)

https://github.com/opensciencegrid/gratia-probe/blob/v2.0.1-2/common/gratia/common/certinfo.py#L140

I guess that means in all cases AuthTokenIssuer maps to both VOName & ReportableVOName.

brianhlin: Ok, in the interest of dropping unnecessary code paths in the future, let's not bother setting the certinfo FQAN:

```suggestion
```

Feel free to add a comment indicating this decision so that we're not confused by the asymmetry in the future

common/gratia/common/condor_ce.py
```diff
@@ -362,9 +367,14 @@ def queryJob(jobid):
...
         certinfo["VO"] = info['x509UserProxyVOName']
+    elif 'AuthTokenIssuer' in info:
+        certinfo["VO"] = info['AuthTokenIssuer']
     if 'x509userproxysubject' in info:
         certinfo['DN'] = info['x509userproxysubject']
+    elif 'AuthTokenSubject' in info:
+        certinfo['DN'] = info['AuthTokenSubject']
     if 'x509UserProxyFirstFQAN' in info:
         certinfo['FQAN'] = info['x509UserProxyFirstFQAN']
+    # intentionally omitting AuthTokenIssuer equivalent for FQAN; see PR #96
```
brianhlin, line 378:
```suggestion
    # FQAN is used as a backup for DN/VO in the certinfo to populate
    # VOName/ReportableVOName in the usage record, respectively.
    # Since DN/VO are always populated in the AuthToken case, we are
    # intentionally omitting AuthTokenIssuer equivalent for FQAN.
    # See https://github.com/opensciencegrid/gratia-probe/pull/96#discussion_r651148685
```

Let's describe the why in addition to the what. Also a direct link to the discussion/comment about dropping this is preferable, especially since there's been a few other discussions in this PR.

brianhlin (approved):
LGTM

PR #108, opened 2021-07-21
Assign exception to separate name (SOFTWARE-4711)

edquist:
Apparently in python3 the `target` of an `except ... as target` statement gets cleared after the except block is hit, so the only way to refer to it later is to assign it to a separate name.

Who knew.

https://docs.python.org/3/reference/compound_stmts.html#except

Reported broken by @LincolnBryant (thanks!)

### DIFF
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -349,7 +349,7 @@ def process_history_file(logfile):
     except IOError as ie:
         DebugPrint(2, "Cannot process %s: (errno=%d) %s" % (logfile, ie.errno,
             ie.strerror))
-        return 0, 0
+        return 0, 0, 0
     added_transient = False
 
     for classad in fd_to_classad(fd):
--- a/condor/condor_meter
+++ b/condor/condor_meter
@@ -315,7 +315,8 @@ def process_history_dirs(dirs):
             e = None
             try:
                 cnt_submit, cnt_found, cnt_alternate = process_history_file(log)
-            except ValueError as e:
+            except ValueError as ex:
+                e = ex
                 DebugPrint(1, "Failed to parse log file: %s\nError was: %s" % (log, e))
                 cnt_submit, cnt_found, cnt_alternate = 0, 0, 0
 


edquist:
Reviewers: assigning you guys largely because this seems to be an instructional bit of wisdom, if (like me) you weren't already aware.

brianhlin:
Yeah, I suppose this works but this feels like hijinks to cover up code smell. `process_history_file()` should probably be catching the `ValueError` and returning `0, 0, 0`.

Separately, this return value is going to result in sadness: https://github.com/opensciencegrid/gratia-probe/blob/60dadd3cab93bc844cc01484a693ed286d12fe8d/condor/condor_meter#L352

edquist:
> Separately, this return value is going to result in sadness

Thanks good catch - i've addressed that now also.

> Yeah, I suppose this works but this feels like hijinks to cover up code smell.

All it is is restoring the python2 behavior the code was intended to have when it was written.

The fact that the existing code smells is exactly why i am touching it as little as possible.

> `process_history_file()` should probably be catching the `ValueError` and returning `(0, 0, 0)`.

That'd be one way to refactor it.  But glancing at process_history_file it's not obvious to me in which place(s) a ValueError might be raised, and i generally hate huge try blocks that span a whole function.  I guess i'd be ok with a wrapper around the original process_history_file that just does the ValueError handling.  But then i would surely need to change the name of this PR ;D

edquist:
> Separately, this return value is going to result in sadness

> it's not obvious to me in which place(s) a ValueError might be raised

You know it now occurs to me the hilarious possibility that that ValueError handling was exactly for assigning `0, 0` to `cnt_submit, cnt_found, cnt_alternate`.

I mean, that would be pretty stellar.  Almost too good to fix.

brianhlin (approved):
> You know it now occurs to me the hilarious possibility that that ValueError handling was exactly for assigning 0, 0 to cnt_submit, cnt_found, cnt_alternate.
>
> I mean, that would be pretty stellar. Almost too good to fix.

Lol, I guess that could happen but I don't think it's why you added it https://opensciencegrid.atlassian.net/browse/SOFTWARE-3877, https://github.com/opensciencegrid/gratia-probe/pull/54

> That'd be one way to refactor it. But glancing at process_history_file it's not obvious to me in which place(s) a ValueError might be raised, and i generally hate huge try blocks that span a whole function. 

Of course, I'd just prefer a specific place to catch this exception but I suspect that `classadToJUR` is raising it and that function is unmanageable. Though we do have the problematic history file so we could figure it out...

> The fact that the existing code smells is exactly why i am touching it as little as possible.

Fair enough and since we plan on tearing a lot of code out of gratia-probe, it's probably not a good use of time to polish potentially dead weight.

edquist:
> Lol, I guess that could happen but I don't think it's why you added it SOFTWARE-3877, #54

Awww... Gross, and no fair using git blame  ;)

I have no good excuse for uglifying the code like that.

Must have been literally asleep at the wheel because this looks like a car wreck.

brianhlin:
Merging. @edquist presumably we'll need this fix for 2.x too, right?

edquist:
We sure do.

edquist:
> Of course, I'd just prefer a specific place to catch this exception but I suspect that `classadToJUR` is raising it and that function is unmanageable.

So (perhaps for future reference) this is discussed in #54.  Apparently `classadToJUR` is not the culprit (if it were, the surrounding code would have already handled a ValueError).  Apparently `classadLib.parseOne()` (called from `fd_to_classad`, called from `process_history_file`) raises a ValueError on a truncated history file.

See [#54 comment 1](https://github.com/opensciencegrid/gratia-probe/pull/54#issuecomment-548487213) and [#54 comment 2](https://github.com/opensciencegrid/gratia-probe/pull/54#issuecomment-551219912).

So it seemed the most specific place to catch this exception without affecting return types was right around the `process_history_file(...)` call, which the existing code does, and which #110 attempts to do less smellily.

Though as i (forgettably) said at the end of #54,

>Yeah the logic was a bit complicated, as you said, and i tried to change as little as possible to add quarantining at the right spot rather than do a major restructuring... But suggestions welcome if you have an idea for a different approach, feel free to make a PR.

brianhlin:
I forgot how much gratia bites back when you try to fix it

PR #109, opened 2021-07-21
Tweaks for gratia.conf (SOFTWARE-4611)

edquist:
- rename `99_` -> `50-` (a lest hostile number, and a less dangerous terminator)
- move the gratia.conf for condor-ce from /etc to /usr/share


edquist:
@brianhlin: i have also renamed the gwms conf file, but i have not changed its location, nor the location of the (now) 50-gratia.conf for the regular condor probe.  Lmk if you wanted those tweaked also.

brianhlin (approved):
The install section confuses me a bit but it at least tracks the previous logic. LGTM

PR #110, opened 2021-07-21
Rework ValueError handling (SOFTWARE-4711)

edquist:
in particular, avoid accessing 'e' outside the except block where it is defined, which doesn't work in python3.

This is an alternative to #108, done instead as a refactor to remove (a bit of) the code smelled by @brianhlin.

Brian, feel free to let me know which of the two you like better.


brianhlin (commented):
I think this is just moving the hackery around so let's just go with #108 :)

PR #112, opened 2021-07-22
create condor_meter.py symlink in test workflow

edquist:
... rather than in the sources

the extra .py file caused .pyc files to be created, and none of these
were being packaged.  we could either rm it from the packaging, or drop
it from the sources and add it to the test workflow, which is what we
are doing here.


brianhlin (approved):
LGTM. Long term I'd like to move more of the Python bits into libraries so we can more easily test them instead of doing this goofy symlink dance

PR #107, opened 2021-07-16
The lack of a log level causes a failure

djw8605:
Traceback:

```  
File "/usr/lib/python3.6/site-packages/gratia/common/send.py", line 60, in Send
    if not XmlChecker.CheckXmlDoc(xmlDoc, False):
  File "/usr/lib/python3.6/site-packages/gratia/common/xml_utils.py", line 55, in CheckXmlDoc
    content = content + checker(xmlDoc, external, resourceType)
  File "/usr/lib/python3.6/site-packages/gratia/common/xml_utils.py", line 218, in UsageCheckXmldoc
    id_info = CheckAndExtendUserIdentity(xmlDoc, userIdentityNodes[0], namespace, prefix, use_certinfo)
  File "/usr/lib/python3.6/site-packages/gratia/common/xml_utils.py", line 439, in CheckAndExtendUserIdentity
    job_certinfo = condor_ce.queryJob(localJobId)
  File "/usr/lib/python3.6/site-packages/gratia/common/condor_ce.py", line 358, in queryJob
    classad['GlobalJobId'])
  File "/usr/lib/python3.6/site-packages/gratia/common/debug.py", line 59, in DebugPrint
    if  level < int(getGratiaConfig().get_DebugLevel()):
TypeError: '<' not supported between instances of 'str' and 'int'
```

PR #103, opened 2021-07-12
Create GHA to run gratia unit tests (SOFTWARE-4684)

mtru32:
Add symlink to satisfy unit tests.

PR #113, opened 2021-07-23
2.x update (SOFTWARE-4708)

edquist:
- Move 99_gratia-ce.conf to /usr/share location (SOFTWARE-4611)
- Fix exception handling in condor_meter (SOFTWARE-4711)
- Fix missing log level (#107)
- GHA workflow (#103)

Eh, the cherry-picking got tedious.

I got curious, and somewhat surprisingly merging `master` into `2.x` worked pretty well.  Really just a handful of manual merge-conflict resolutions in the spec file.  The Commits list for this PR may look long but you can make sense of everything easily enough in the Files changed tab.


brianhlin:
I think somehow the merge missed removal of the symlink? https://github.com/opensciencegrid/gratia-probe/pull/113/checks?check_run_id=3146571563#step:4:4

edquist:
I'm actually happy enough with how this went that i would not mind in the future doing merges of master back into 2.x as PRs (with whatever merge-conflict resolution might be necessary, mostly in the spec or else skipping files that are deleted in 2.x), rather than making the same changes against 2.x.

I'm open to having better sense talked into me, but the "wait do i have the changes from this commit?" game gets old kind of fast if there is a backlog.

(And just to note it, somehow a rebase resulted in a lot more merge conflicts than a regular merge - basically at every commit, probably because there was a mix of some changes that were already applied in 2.x.)

edquist:
> I think somehow the merge missed removal of the symlink?

:fearful:

Well that's disappointing, git.  Why would you play with my trust like that?

Good catch, it's removed now.  No idea why the merge didn't take it....

brianhlin (approved):
LGTM!

PR #117, opened 2021-08-20
update default ProbeConfig directories (SOFTWARE-4621)

edquist:
Per discussion in SOFTWARE-4596


brianhlin:
@edquist wouldn't this change the default for the payload accounting case, too? I think we want to avoid that. Also, it'd be good to drop comments into the config so that we know what gratia needs these dirs for so that we don't have to reconstruct their meaning every time.

brianhlin:
Closing this in favor of #119

PR #118, opened 2021-08-24
Update dCache-transfer probe and common2 lib for Python 3

retzkek:
The hotshot profiler was removed from from python 3, so I just stripped out the built-in profiling. If you want to profile a script now, you can just run `python -m cProfile myscript.py` (see https://docs.python.org/3/library/profile.html).

I'm still figuring out how to test this, since I don't have a test dCache environment or database. I may have to build a pre-release RPM (1.24.1-0.1?) and have our dCache admins do it live.

### DIFF
--- a/common2/gratia/common2/checkpoint.py
+++ b/common2/gratia/common2/checkpoint.py
@@ -280,7 +280,7 @@ class DateTransactionCheckpoint(Checkpoint):
 
     def _load(self, target):
         pkl_file = open(target, 'rb')
-        self._dateStamp, self._transaction = pickle.load(pkl_file)
+        self._dateStamp, self._transaction = pickle.load(pkl_file,encoding='latin1')
         pkl_file.close()
 
     def get_val(self):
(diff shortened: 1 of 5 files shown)


retzkek:
I've pushed a couple more python 2 to 3 errors caught in testing. Let's give ops a couple more days to shake the limbs.

retzkek:
OK I think this is good to go.

edquist:
Thanks @retzkek - and sorry you had to deal with this.

So this works for you now?

Admittedly the common2 stuff does not get any love or testing from us, so if it works for you that's probably good enough for me.

retzkek:
> So this works for you now?

Yep, probes seem to be running fine now. :knock:

> Admittedly the common2 stuff does not get any love or testing from us

No worries, since we're the only user there's no real reason for OSG to worry about them. Just a matter of time before they get retired.

edquist (approved):
Thanks and Godspeed

brianhlin:
@edquist could you merge these changes into the 2.x branch?

edquist:
@brianhlin - done!

PR #119, opened 2021-09-04
Improve ProbeConfig generation for 2.x (SOFTWARE-4621)

edquist:
@brianhlin was kind enough to let me / trick me into spending my friday afternoon (and evening) improving the packing, especially as it relates to generating ProbeConfigs with per-probe customization.  Maybe this will even inspire making further improvements, but for now this PR will do the following:

- move some ugly probe-specific sed customizations out of rpm spec into ProbeConfig.add
- add a new script for merging ProbeConfig.add files, which allows updating existing attributes
- move hacky htcondor-ce folder customizations into ProbeConfig.add
- make htcondor-ce probe a bit less special by giving it its own directory for its probe-specific files
- pull in a few potentially interesting comments from ProbeConfigTemplate into the .osg version, and drop the former
- make a few incidental packaging cleanups


common/update-probeconfig.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def main(pc, pc_add, pc_out=None):
+    add_txt = read_if_exists(pc_add)
+    pc_txt = open(pc).read().replace('@PROBE_SPECIFIC_DATA@', add_txt)
+
+    lines = reversed(pc_txt.splitlines())
+    lines = list(filter_nonlast_lines(lines))
+    with open(pc_out or pc, "w") as outf:
+        print('\n'.join(reversed(lines)), file=outf)
```
brianhlin, line 37: wat. Comment here would be helpful

htcondor-ce/ProbeConfig.add
```diff
@@ -0,0 +1,11 @@
+
+    DataFolder="/var/lib/condor-ce/gratia/data"
+    WorkingFolder="/var/lib/condor-ce"
+    LogFolder="/var/log/condor-ce/gratia"
```
brianhlin, line 4: We should create this dir in the packaging

brianhlin (approved):
Looks sensible, which is high praise for Gratia. A few comments but pre-approving

common/update-probeconfig.py
edquist: > wat. Comment here would be helpful

"Left as an exercise to the reader."

Wat, didn't you see the title of this book?

_Python Puzzles_
_Comments Confuse_
_Lambdas Lament_
_Unless You Defuse_

edquist: I've added comments and rejiggered this a bit to attempt to improve readability.

htcondor-ce/ProbeConfig.add
edquist: Right!  Adding...

common/update-probeconfig.py
edquist: @brianhlin i know you've already pre-approved, but i'm throwing this back to you just to see if you are satisfied with the comments / (hopefully improved) readability updates for [update-probeconfig.py](https://github.com/edquist/gratia-probe/blob/SOFTWARE-4621.ProbeConfig-cleanup/common/update-probeconfig.py).

brianhlin: Yeah, that helps. Thanks!

PR #121, opened 2021-09-27
Update htcondor-ce WorkingFolder to match hosted-ce container (SOFTWARE-4806)

edquist:
The hosted-ce container sets WorkingFolder to /var/lib/condor-ce/gratia/tmp.  Create and package that dir for the htcondor-ce probe, and set WorkingFolder to it in that probe's config.


brianhlin (commented):
Do we really want these dirs to be chmod `1777`? Seems overly permissive to me

brianhlin:
While you're in there, can you fix the permissions of `/var/lib/condor-ce/gratia/data`?

brianhlin (approved):
LGTM

PR #123, opened 2021-10-06
Consolidate condor/gwms probes into condor-ap (SOFTWARE-4846)

edquist:
- rename condor probe to condor-ap
- update package name and paths to match in rpm spec
- pull in gwms config from old glideinwms probe
- add upgrade path from condor & glideinwms probes

### DIFF
--- a/rpm/gratia-probe.spec
+++ b/rpm/gratia-probe.spec
@@ -194,9 +194,7 @@ git_commit_id=$(gzip -d < %{SOURCE0} | git get-tar-commit-id)
 
   # Install condor configuration snippet
   install -d $RPM_BUILD_ROOT/%{_datadir}/condor/config.d
-  install -m 644 condor-ap/50-gratia.conf $RPM_BUILD_ROOT/%{_datadir}/condor/config.d/50-gratia.conf
   install -m 644 condor-ap/50-gratia-gwms.conf $RPM_BUILD_ROOT/%{_datadir}/condor/config.d/50-gratia-gwms.conf
-  rm $RPM_BUILD_ROOT%{_datadir}/gratia/condor-ap/50-gratia.conf
   rm $RPM_BUILD_ROOT%{_datadir}/gratia/condor-ap/50-gratia-gwms.conf
 
   # Install the htcondor-ce configuration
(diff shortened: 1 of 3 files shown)


rpm/gratia-probe.spec
```diff
@@ -310,27 +312,32 @@ fi
...
-%doc %{default_prefix}/gratia/condor/README
-%dir %{default_prefix}/gratia/condor
-%{default_prefix}/gratia/condor/condor_meter
+%doc %{default_prefix}/gratia/condor-ap/README
+%dir %{default_prefix}/gratia/condor-ap
+%{default_prefix}/gratia/condor-ap/condor_meter
 %config(noreplace) %{_sysconfdir}/condor/config.d/50-gratia.conf
-%config(noreplace) %verify(not md5 size mtime) %{_sysconfdir}/gratia/condor/ProbeConfig
-%config(noreplace) %{_sysconfdir}/cron.d/gratia-probe-condor.cron
+%config(noreplace) %{_sysconfdir}/condor/config.d/50-gratia-gwms.conf
```
brianhlin, line 335: I prefer this config living in  `/usr/share/condor/config.d/` as `%config`

condor-ap/gratia-probe-condor-ap.cron
```diff
@@ -0,0 +1 @@
+0,15,30,45 * * * * root /usr/share/gratia/common/cron_check /etc/gratia/condor-ap/ProbeConfig && /usr/share/gratia/condor-ap/condor_meter -s 900
```
brianhlin, line 1: We can put this in a separate PR if you prefer but I'd really like to go the `SCHEDD_CRON` route and it appears that @rynge  is on board: https://github.com/opensciencegrid/osg-flock/blob/master/rpm/80-osg-flocking.conf#L56-L62

rpm/gratia-probe.spec
```diff
@@ -310,27 +312,32 @@ fi
...
+%config(noreplace) %verify(not md5 size mtime) %{_sysconfdir}/gratia/condor-ap/ProbeConfig
```
brianhlin, line 336: It'd be good to compare what we ship vs what @rynge actually wants https://github.com/opensciencegrid/osg-flock/blob/master/rpm/ProbeConfig-OSPool.

Though I imagine that we'll want to change a bunch of the dirs to accommodate a `SCHEDD_CRON` like what we have to do for the CE probe.

brianhlin (commented):
See specific comments

rpm/gratia-probe.spec
```diff
@@ -310,27 +312,32 @@ fi
...
-%dir %{default_prefix}/gratia/condor
-%{default_prefix}/gratia/condor/condor_meter
-%config(noreplace) %{_sysconfdir}/condor/config.d/50-gratia.conf
-%config(noreplace) %verify(not md5 size mtime) %{_sysconfdir}/gratia/condor/ProbeConfig
-%config(noreplace) %{_sysconfdir}/cron.d/gratia-probe-condor.cron
+%doc %{default_prefix}/gratia/condor-ap/README
+%dir %{default_prefix}/gratia/condor-ap
+%{default_prefix}/gratia/condor-ap/condor_meter
+%config %{_datadir}/condor/config.d/50-gratia.conf
+%config %{_datadir}/condor/config.d/50-gratia-gwms.conf
```
brianhlin, line 335: We may as well combine these into a single config file

PR #130, opened 2021-12-09
Collect container startd ads from all OSPool collectors (SOFTWARE-4928)

brianhlin:
The weird line breaks in the startd hash function are because things were starting to look real lispy


matyasselmeci (approved):
Seems reasonable

PR #134, opened 2021-12-13
Fixup for SOFTWARE-4928 and add some logging on query failure

brianhlin:


brianhlin:
Oops, something still looks off

brianhlin:
I dropped in the file from my branch into the container and things seem appear to be chugging along!

matyasselmeci:
What happens to an upgrade install if there's a database present with the old schema?

brianhlin:
I imagine that it'd fail! But I moved the old db out of the way so it's running with the new schema so we don't have to worry about it this time around

matyasselmeci (approved):

PR #136, opened 2022-01-11
Make client authN optional; necessary for OSPool collector queries

brianhlin:
Tested this manually in the container and this fixed queries against `cm-1` and `cm-2`


matyasselmeci:
What kind of queries are these? condor_q or condor_status? Can't the queries use the container's token?

matyasselmeci:
Let's try and come up with a way to make this work without disabling INTEGRITY and ENCRYPTION; if we can't, or if it will take software or HTCondor work, then we can use this as a temporary fix in the meantime.

brianhlin:
`condor_status` queries to get the ads of the backfill container slots. We could get a READ token for user but the OSPool collector doesn't require one as it's open to the world for READ. This config just overrides the "secure-by-default" configuration in the newer versions of HTCondor.

brianhlin:
IIRC, integrity and encryption forces authentication so we'd have to get the token. I can poll the HTCSS team about this but I'd prefer at least getting a container that will hold us over instead of my hand-placed config in the existing container

brianhlin:
We fixed the certs on the HA collectors so we don't need this anymore

PR #137, opened 2022-01-25
Fix quarantine dir ownership (SOFTWARE-4975)

brianhlin:


rpm/gratia-probe.spec
```diff
@@ -407,6 +408,7 @@ The HTCondor-CE probe for the Gratia OSG accounting system.
 %doc %{default_prefix}/gratia/htcondor-ce/README
 %{default_prefix}/gratia/htcondor-ce/condor_meter
 %attr(0755,condor,condor) %dir %{_sharedstatedir}/condor-ce/gratia/data
+%attr(0755,condor,condor) %dir %{_sharedstatedir}/condor/gratia/data/quarantine
```
matyasselmeci:
```suggestion
%attr(0755,condor,condor) %dir %{_sharedstatedir}/condor-ce/gratia/data/quarantine
```
?

matyasselmeci (approved):
One question, otherwise pre-approving.

rpm/gratia-probe.spec
brianhlin: Right! copy-paste fail :(

PR #138, opened 2022-01-26
Add gratia-probe-condor-batch (SOFTWARE-4978)

brianhlin:


rpm/gratia-probe.spec
```diff
@@ -400,7 +436,7 @@ Requires: htcondor-ce
 License: See LICENSE.
 
 %description htcondor-ce
-The HTCondor-CE probe for the Gratia OSG accounting system.
+The HTCondor-CE probe for the Gratia OSG accounting system. Designed for use with non-HTCondor batch systems.
```
matyasselmeci, line 442: This suggests we don't want it running if the batch system is condor?

rpm/gratia-probe.spec
```diff
@@ -287,19 +295,47 @@ fi
...
 # %%{default_prefix}/gratia/common2/probeinput.py
 
+%package condor-batch
+Summary: A probe accounting for pilot jobs in a local HTCondor batch system
+Requires: %{name}-common = %{version}-%{release}
+Requires: condor
+Requires: %{condor_python}
+Requires: htcondor-ce
+Provides: %{name}-condor
+Obsoletes: %{name}-condor
```
matyasselmeci:
```suggestion
Provides: %{name}-condor = %{version}-%{release}
Obsoletes: %{name}-condor < 2.4.0
```

rpm/gratia-probe.spec
brianhlin: Yeah, config and crons would fight

matyasselmeci: wanna add a Conflict?

brianhlin: Done in the latest commit

matyasselmeci (approved):
LGTM.

PR #145, opened 2022-04-21
fresh tags are deprecated, let's give 3.6 EL8 a whirl

brianhlin:
*-release tags contain the 'osg' and 'osg-upcoming' repos


edquist (approved):
looks straightforward enough

PR #147, opened 2022-05-10
downgrade certinfo errors/warnings to debug (SOFTWARE-4980)

edquist:
@brianhlin per discussion this morning as a first step for SOFTWARE-4980 we want to eliminate the certinfo error and warning messages.

I wasn't sure if it would be best to remove them entirely, so I've just downgraded them to debug-level messages.

Let me know if you'd like to handle this differently.

### DIFF
--- a/common/gratia/common/certinfo.py
+++ b/common/gratia/common/certinfo.py
@@ -184,9 +184,9 @@ def readCertInfoLog(localJobId):
                     else:
                         res['FQAN'] = None
                     res['VO'] = None
-                    DebugPrint(0, 'Warning: found valid certinfo file for '+str(localJobId)+' in the log files: ' + pattern + ' with ' + str(res))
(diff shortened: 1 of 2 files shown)


brianhlin:
When we're troubleshooting, we generally tell people to set the DebugLevel to 5, don't we? I'd rather not have these messages polluting our troubleshooting session so I'd rather remove them outright

edquist:
Alright, i've updated the PR to comment these out instead.  Effectively they are removed, but i'm leaving in the comments to help make sense of the surrounding certinfo code for when we rip all that out too.

brianhlin:
Fine by me. Could you add a comment explaining why we're commenting them out with a reference to the Jira ticket?

edquist:
Sure thing; added.

brianhlin (approved):
LGTM

PR #148, opened 2022-05-16
Replace AuthToken with routed job attrs (SOFTWARE-5185)

edquist:
Per SOFTWARE-5185, the AuthToken* attributes we used for VO info in the
htcondor-ce probe don't exist in the routed job for non-condor batch
systems.  And, even worse, these may have entirely different values for
condor batch systems (a different token can be used to authN to the local
schedd).

In HTCONDOR-1071, htcondor-ce config is added to copy the pre-routed
job's AuthToken* attrs to orig_AuthToken* attrs in the routed job.

Thus, SOFTWARE-5185 calls for replacing the AuthToken* references in the
htcondor-ce probe with orig_AuthToken*.  In order to make this
probe-specific, I add a probe_name check, as the condor-ap and
htcondor-ce probes share the same probe code.

Note also that AuthToken* references exist in common/condor_ce.py,
in certinfo code (which seems to be unimportant at this point, since
it's going away), but also in CheckAndExtendUserIdentity from
common/xml_utils.py, to fill in missing vo info for jobs by querying the
htcondor-ce directly.  Do these need to get the "orig_" prefix also?
(Always? Just when running the htcondor-ce probe?  Never?)

...

It _seems_ to me that for the condor_ce.py code we just need to update the attr names in the condor_ce_q query, which is made against routed jobs - and that being the case i always query the `orig_`-prefixed names (not just in the htcondor-ce probe case).

### DIFF
--- a/common/gratia/common/condor_ce.py
+++ b/common/gratia/common/condor_ce.py
@@ -304,8 +304,8 @@ condor_ce_q -const 'RoutedJob =?= true && GridJobId =!= UNDEFINED \
     -format 'x509UserProxyVOName=%s\t' x509UserProxyVOName \
     -format 'x509UserProxyFirstFQAN=%s\t' x509UserProxyFirstFQAN \
     -format 'x509userproxysubject=%s\t' x509userproxysubject \
-    -format 'AuthTokenIssuer=%s\t' AuthTokenIssuer \
-    -format 'AuthTokenSubject=%s\t' AuthTokenSubject \
+    -format 'AuthTokenIssuer=%s\t' orig_AuthTokenIssuer \
+    -format 'AuthTokenSubject=%s\t' orig_AuthTokenSubject \
     -format 'GridJobId=%s' GridJobId \
     -format '\n' junk \
 """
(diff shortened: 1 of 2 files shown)


brianhlin (approved):
> Thus, SOFTWARE-5185 calls for replacing the AuthToken* references in the
htcondor-ce probe with orig_AuthToken*. In order to make this
probe-specific, I add a probe_name check, as the condor-ap and
htcondor-ce probes share the same probe code.

This is fine, though I don't think that the `AuthToken*` attrs will have any semantic meaning for the `condor-ap` probe, at least not yet.

> Note also that AuthToken* references exist in common/condor_ce.py,
in certinfo code (which seems to be unimportant at this point, since
it's going away), but also in CheckAndExtendUserIdentity from
common/xml_utils.py, to fill in missing vo info for jobs by querying the
htcondor-ce directly. Do these need to get the "orig_" prefix also?
(Always? Just when running the htcondor-ce probe? Never?)

I say we change all of the references, just in case. I don't recall what `CheckAndExtendUserIdentity` hooks into but we certainly only added the `AuthToken*` attrs as a replacement for the `x509userproxy*` attrs for extracting VO data.

Otherwise, LGTM, pre-approving

edquist:
> I don't think that the `AuthToken*` attrs will have any semantic meaning for the `condor-ap` probe, at least not yet.

If that's the case, i can remove the conditionalization and always use the `orig_AuthToken*` names, regardless of the `probe_name` (that is, keep it for condor-ap).

> I say we change all of the references, just in case

Okie doke.

PR #150, opened 2022-05-27
fix wrongful tuple concatenation (SOFTWARE-5199)

edquist:
responseString is a str, never made sense to add a tuple to it.

this dates back over 15 years, to 17d8465b, where presumably a comma was accidentally used instead of a plus.  apparently it didn't get seen much, as it only gets tripped on send failure.


brianhlin (approved):
LGTM

PR #152, opened 2022-06-22
Add back services subpackage to 2.x (SOFTWARE-5236) 

edquist:
Still required by the storage probes


brianhlin (approved):
Hrm, an interesting probe. Looks reasonable to me

edquist:
I don't think it's a probe exactly; i think it's like common code used for the storage probes

edquist:
Ho boy, i guess it is a probe?  Now i am confused too! :)

PR #154, opened 2022-06-29
Fixup broken dcache-transfer probe (SOFTWARE-5241)

edquist:
**1. fix global terminationAlarm handling**

I was getting an UnboundLocalError exception (local variable
'terminationAlarm' referenced before assignment), because main
sets terminationAlarm in the big try block, without calling it
global, which makes it a local variable for the whole function.

that actually means the global wasn't getting set, which is a bug,
but we happened to see it because the test later in main for
terminationAlarm references the expected-to-be-local copy in main,
which happened not to be set because we didn't make it that far
in the try block before there was an exception (apparently).

While we're at it, i'm switching the "!= None" check to the always
preferred "is not None" check.

...

**2. restore /usr/share ProbeConfig symlinks**

This reverts 4326a64.

Apparently the /usr/share ProbeConfig symlinks were actually used by
some of the storage/transfer probes in order to refer to a default
ProbeConfig location, which was the working dir, which in the case
of these probes was the /usr/share location for the probe script
itself.  That is, the ProbeConfiguration system by default looks
for "ProbeConfig" in the current working directory, in order to
pick up the symlink to the real file under /etc.

This was not needed for the condor probes, which provide an explicit
path for the ProbeConfig; but some of the less-well-tested in 2.x
probes (basically just used by FNAL) apparently relied on the
symlink.

Now that i understand what they were for, for consistency i'm adding
the /usr/share ProbeConfig symlink to the condor probes as well, even
though they don't rely on them.

**3. add unexpected exception handling**

The dcache-transfer probe was sending tracebacks for unexpected exceptions to /dev/null, which was super unhelpful for debugging.

### DIFF
--- a/rpm/gratia-probe.spec
+++ b/rpm/gratia-probe.spec
@@ -128,6 +128,8 @@ git_commit_id=$(gzip -d < %{SOURCE0} | git get-tar-commit-id)
     PROBE_DIR=$RPM_BUILD_ROOT/%{_sysconfdir}/gratia/$probe
     install -d $PROBE_DIR
     install -m 644 common/ProbeConfigTemplate.osg $PROBE_DIR/ProbeConfig
+    ln -s %{_sysconfdir}/gratia/$probe/ProbeConfig \
+          $RPM_BUILD_ROOT/%{_datadir}/gratia/$probe/ProbeConfig
 
     ## Probe-specific customizations
     # Replace @PROBE_SPECIFIC_DATA@ with ProbeConfig.add (if any)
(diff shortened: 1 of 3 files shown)


edquist:
Note from the support ticket (https://support.opensciencegrid.org/a/tickets/69921), dropping in the ProbeConfig symlinks fixed the issue for Yujun.  The other items addressed in this PR are not strictly necessary i guess, but i think they are worthy incidental fixes.

@brianhlin / @retzkek, if somebody can give this a review then I can get this into a new build for Yujun ...

rpm/gratia-probe.spec
```diff
@@ -128,6 +128,8 @@ git_commit_id=$(gzip -d < %{SOURCE0} | git get-tar-commit-id)
     PROBE_DIR=$RPM_BUILD_ROOT/%{_sysconfdir}/gratia/$probe
     install -d $PROBE_DIR
     install -m 644 common/ProbeConfigTemplate.osg $PROBE_DIR/ProbeConfig
+    ln -s %{_sysconfdir}/gratia/$probe/ProbeConfig \
+          $RPM_BUILD_ROOT/%{_datadir}/gratia/$probe/ProbeConfig
```
brianhlin, line 130: Could you add a short comment here explaining why?

brianhlin (approved):
LGTM except for a minor comment. At some point we should change `%{default_prefix}` to `%{_datadir}`

rpm/gratia-probe.spec
edquist: Done

edquist:
> At some point we should change `%{default_prefix}` to `%{_datadir}`

Done.  Apparently  `%{default_prefix}`, along with the `Prefix:`'es, were put there in the long long ago to allow for relocatable packages (eg putting things under `/opt` by running rpmbuild with a `--prefix` arg).  But that is not a thing we care about so i've removed all that.

PR #156, opened 2022-07-11
tweak ResourceName calculation (SOFTWARE-5249)

edquist:
we're doing two things here:

1. add GLIDEIN_ResourceName to SYSTEM_JOB_MACHINE_ATTRS config

2. Prefer MachineAttrGLIDEIN_ResourceName0 over
   MATCH_EXP_JOBGLIDEIN_ResourceName

Per SOFTWARE-5249's description:

This will cause the AP to stuff an attribute into the job ad for each
match to a slot with the format MachineAttrGLIDEIN_ResourceNameN where
MachineAttrGLIDEIN_ResourceName0 is the slot's GLIDEIN_ResourceName for
the last job-slot match. So additionally, we need to update condor_meter
to prefer MachineAttrGLIDEIN_ResourceName0 over
MATCH_EXP_JOBGLIDEIN_ResourceName.

-----

@brianhlin, can you confirm whether i should simply add the `SYSTEM_JOB_MACHINE_ATTRS` line to the config, or whether this should replace the above `JOBGLIDEIN_ResourceName` definition?

### DIFF
--- a/condor-ap/50-gratia-gwms.conf
+++ b/condor-ap/50-gratia-gwms.conf
@@ -21,6 +21,11 @@ JOBGLIDEIN_ResourceName="$$([IfThenElse(IsUndefined(TARGET.GLIDEIN_ResourceName)
 # this host
 SUBMIT_ATTRS = $(SUBMIT_ATTRS) JOBGLIDEIN_ResourceName
 
+# Have AP stuff an attribute into the job ad for each match to a slot
+# with the format MachineAttrGLIDEIN_ResourceNameN where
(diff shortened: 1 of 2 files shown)


brianhlin:
> @brianhlin, can you confirm whether i should simply add the SYSTEM_JOB_MACHINE_ATTRS line to the config, or whether this should replace the above JOBGLIDEIN_ResourceName definition?

Add! Let's carry both for now

edquist:
>Add! Let's carry both for now

Cool; that's what the PR does currently.

brianhlin (approved):
LGTM!

PR #159, opened 2022-08-01
apply storage probe fixes from Yujun (SOFTWARE-5279)

edquist:
See https://support.opensciencegrid.org/a/tickets/69921

...

Apparently there was a nesting issue with `user_space` ?

### DIFF
--- a/enstore-storage/enstore-storage
+++ b/enstore-storage/enstore-storage
@@ -253,6 +253,8 @@ date cannot be null.
         #end_time = timeutil.at_minute(datetime.datetime.now())
         end_time = None
         # Give to the end time a 60 seconds delay. To allow for enstore DB processing time
+        if limit is None:
+            limit = 0
         if limit > 0:
             end_time = timeutil.wind_time(start_time, hours=limit, backward=False)
             end_time = min(end_time, timeutil.wind_time(datetime.datetime.now(), seconds=60))
--- a/dCache-storagegroup/dcache-storagegroup
+++ b/dCache-storagegroup/dcache-storagegroup
@@ -63,11 +63,10 @@ def sendRecords(fn):
     totals_by_sg={}
     for pg,item1 in user_space.items():
         storage_groups[pg] = {}
-        for pool, item2 in item1.items():
-            for sg, value in item2.items():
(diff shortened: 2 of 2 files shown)


enstore-storage/enstore-storage
```diff
@@ -253,6 +253,8 @@ date cannot be null.
         #end_time = timeutil.at_minute(datetime.datetime.now())
         end_time = None
         # Give to the end time a 60 seconds delay. To allow for enstore DB processing time
+        if limit is None:
+            limit = 0
         if limit > 0:
```
brianhlin, line 258:
```suggestion
        elif limit > 0:
```

No reason to check this if it's originally `None` and we explicitly set it to 0

retzkek:
I need to get some more info from Yujun before I can review this.

edquist:
Thanks @retzkek.  Yeah notably in the support ticket Yujun said about his changes, "Two probes had some python2/python3 related issues. I fixed the code locally:" ... But Yujun's suggested fix does not seem to be a python2/3 issue, it looks more like he thinks the structure of the pickled object getting passed in is different than the (old) code expects.

EDIT: Ah yeah I see you asked in the support ticket about the data structure written to the tmp file.

enstore-storage/enstore-storage
edquist: That's true but changing this to an `elif` like you have it would actually break the logic he was trying to fix, as it would prevent the `else` block a few lines down from firing when `limit` _is_ `None`.

edquist, line 258: But i think we can still clean it up a bit with

```suggestion
        if limit is not None and limit > 0:
```

or just

```suggestion
        if limit and limit > 0:
```

...

Or if we happen to know that limit is not negative (I do not know whether that is a safe assumption), just

```suggestion
        if limit:
```

edquist: PS i think this spot must have been the python2/3 issue Yujun had in mind, as `None > 0` worked as a comparison in python2 but raises an exception in python3 (which is why he added this explicit None handling).

brianhlin:
There doesn't seem to be an impetus for this any longer. Abandoning.

PR #160, opened 2022-08-05
Fix condor-ap probe bug(s) in resource name detection (SOFTWARE-5285)

edquist:
regarding the fix in `get_classad_resource_name()`, David Schultz reported in icecube-condor:

> we have at least one site where the first value is undefined while the second is correct. those are getting put into quarantine right now. maybe there should be a check that the attr is valid?

...

And the second bit that had `host_descr = classad[attr]` sadly was just broken. Incomplete refactoring when i broke out `get_classad_resource_name()` into a separate function.


djw8605 (approved):
Small change, looks good to me.

PR #162, opened 2023-01-09
store LastRemotePool of payload jobs (SOFTWARE-5438)

edquist:
My first whack at this...

get_collector_host() falls back on `condor_config_val COLLECTOR_HOST`, which is mostly borrowed from htcondor_configured()

### DIFF
--- a/condor-ap/condor_meter
+++ b/condor-ap/condor_meter
@@ -782,6 +782,11 @@ def classadToJUR(classad):
             r.Network(val, storageUnit='b', phaseUnit=networkPhaseUnit, metric=attr, description=attr)
     r.Network(total_network, storageUnit='b', phaseUnit=networkPhaseUnit, metric="total")
 
+    if not setIfExists(r.ExecutePool, classad, 'LastRemotePool', "Pool Host"):
+        collector_host = get_collector_host()
+        if collector_host:
+            r.ExecutePool(collector_host, "Pool Host from COLLECTOR_HOST")
+
     # Additional Pegasus attributes
     if 'pegasus_root_wf_uuid' in classad:
         r.AdditionalInfo("PegasusRootWFUUID", classad['pegasus_root_wf_uuid'])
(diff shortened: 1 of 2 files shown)


edquist:
oddly the CI failure is erroring on the `import classad` in the test module

timtheisen:
There was a problem with the HTCondor Python bindings in Pypi. I uploaded a corrected version to Pypi.

condor-ap/condor_meter
```diff
@@ -893,6 +898,49 @@ def htcondor_configured():
...
+            cmd = subprocess.Popen(args, stdout=subprocess.PIPE)
+        else:
+            # args needs to be string when shell=True
+            args = condor_config_binary + ' ' + ' '.join(condor_config_val_args)
+            cmd = subprocess.Popen(args, stdout=subprocess.PIPE, shell=True)
+
+        cmd_stdout = cmd_stdout.strip()
+        if cmd.returncode != 0:
+            DebugPrint(-1, "WARNING: condor_config_val COLLECTOR_HOST "
+                           "returned a non-zero return code. Maybe the schedd "
```
brianhlin:
```suggestion
                           "returned a non-zero return code. Maybe the schedd "
```

I think `condor_config_val` queries the `condor_master` daemon by default

condor-ap/condor_meter
```diff
@@ -893,6 +898,49 @@ def htcondor_configured():
...
+                           "is overloaded.")
+            return False
+        elif 'Not defined' in cmd_stdout:
+            DebugPrint(-1, "WARNING: COLLECTOR_HOST not set in the default "
+                           "condor schedd config. You may need to change or "
```
brianhlin: Same thing about `condor_master`

condor-ap/condor_meter
```diff
@@ -893,6 +898,49 @@ def htcondor_configured():
...
+                           "reload the condor configuration or specify a "
+                           "different CondorScheddName in the Gratia config.")
```
brianhlin:
```suggestion
                           "different CondorScheddName in the Gratia config.")
```

Hrm, I don't see where `CondorScheddName` comes into play here

condor-ap/condor_meter
```diff
@@ -782,6 +782,11 @@ def classadToJUR(classad):
             r.Network(val, storageUnit='b', phaseUnit=networkPhaseUnit, metric=attr, description=attr)
     r.Network(total_network, storageUnit='b', phaseUnit=networkPhaseUnit, metric="total")
 
+    if not setIfExists(r.ExecutePool, classad, 'LastRemotePool', "Pool Host"):
+        collector_host = get_collector_host()
+        if collector_host:
+            r.ExecutePool(collector_host, "Pool Host from COLLECTOR_HOST")
```
brianhlin: Can we do some massaging here to handle

1. Multiple hosts
2. Sinful strings
3. Hosts with ports

I think this field should just be an XML list (as you suggsted) of FQDNS

condor-ap/condor_meter
edquist: ah, good catch

condor-ap/condor_meter
edquist: yeah this text was mostly copied from another function, apparently i only looked closely enough to update the referenced COLLECTOR_HOST attribute name

condor-ap/condor_meter
edquist: Sure thing!

condor-ap/condor_meter
edquist: Updated

condor-ap/condor_meter
edquist: Updated - dropped the reference to the CondorScheddName knob

condor-ap/condor_meter
edquist: Updated

condor-ap/condor_meter
edquist: Ok I believe my latest push should handle all these cases, and include one or more `<ExecutePool>` tags in the XML

edquist:
Ping @brianhlin, @djw8605 - I believe I addressed Brian's comments above and I don't think I was planning to add anything else here.  If we still want this feature the PR still needs a review to sign off on it.

condor-ap/condor_meter
```diff
@@ -893,6 +897,68 @@ def htcondor_configured():
...
+          return None
+        else:
+          condor_config_binary = 'condor_config_val'
+        condor_config_path = os.path.join(path, "bin", condor_config_binary)
+        condor_config_val_args = ['COLLECTOR_HOST']
+
+        if (os.path.exists(condor_config_path) and
+            os.path.isfile(condor_config_path)):
+            args = [condor_config_path] + condor_config_val_args
+            cmd = subprocess.Popen(args, stdout=subprocess.PIPE)
```
djw8605, line 917: Aren't we requiring htcondor python bindings already.  Why not use the python bindings to the the config value?

djw8605 (commented):
Just 1 question.

condor-ap/condor_meter
edquist: That's a fair question - I am just following suit with the other parts of the probe that use the CLI tools.

djw8605: I vote using the python API.

brianhlin: @djw8605 I agree but I'd prefer that we address that in a separate PR

djw8605: It can be iterative.  Use python API for this call now.  And a separate PR to fix everywhere else...

brianhlin: Ya, I'd rather get this in. We'll tackle the CLI vs python bindings usage afterwards

brianhlin (approved):
LGTM

PR #164, opened 2023-02-10
use %python3_sitelib to fix EL9 build (SOFTWARE-5416)

edquist:
%python_sitelib apparently does not exist for el9.

Note also that python3-devel is apparently needed in el9 in order for %python3_sitelib to get defined, so I needed to add that, too.


brianhlin (approved):
LGTM

PR #168, opened 2023-03-08
Fix var name in get_collector_host_list()

matyasselmeci:
and handle `host` not being a host:port (not 100% sure it's necessary but better safe than sorry)


brianhlin (approved):
LGTM

PR #169, opened 2023-03-13
Add logging when there are failures to create sandbox directories

djw8605:
This can happen if directory ownership is incorrect.

### DIFF
--- a/common/gratia/common/sandbox_mgmt.py
+++ b/common/gratia/common/sandbox_mgmt.py
@@ -569,8 +569,8 @@ def OpenNewRecordFile(dirIndex):
             if not os.path.exists(working_dir):
                 try:
                     Mkdir(working_dir)
-                except Exception as e:
-                    DebugPrint(0, 'Warning: Exception caught while creating directory: ' + working_dir + ':' + e)
+                except Exception as exc:
+                    DebugPrint(0, 'Warning: Exception caught while creating directory: ' + working_dir + ':' + exc)
                     continue
             if not os.path.exists(working_dir):
                 DebugPrint(0, 'Warning: Directory does not exist even after attempting a Mkdir: ' + working_dir)
--- a/common/gratia/common/sandbox_mgmt.py
+++ b/common/gratia/common/sandbox_mgmt.py
@@ -569,11 +569,14 @@ def OpenNewRecordFile(dirIndex):
             if not os.path.exists(working_dir):
                 try:
                     Mkdir(working_dir)
-                except:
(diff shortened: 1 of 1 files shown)


djw8605:
Untested!

common/gratia/common/sandbox_mgmt.py
```diff
@@ -569,11 +569,14 @@ def OpenNewRecordFile(dirIndex):
             if not os.path.exists(working_dir):
                 try:
                     Mkdir(working_dir)
-                except:
+                except Exception as e:
+                    DebugPrint(0, 'Warning: Exception caught while creating directory: ' + working_dir + ':' + e)
```
brianhlin:
```suggestion
                except Exception as exc:
                    DebugPrint(0, 'Warning: Exception caught while creating directory: ' + working_dir + ':' + exc)
```

I said I was going to be annoying about PEP8

common/gratia/common/sandbox_mgmt.py
```diff
@@ -582,7 +585,9 @@ def OpenNewRecordFile(dirIndex):
                 f = open(filename, 'w')
                 dirIndex = index
                 return (f, dirIndex)
-            except:
+            except Exception as e:
+                DebugPrint(0, 'Caught exception while creating file: ', e)
```
brianhlin:
```suggestion
            except Exception as exc:
                DebugPrint(0, 'Caught exception while creating file: ', exc)
```

common/gratia/common/sandbox_mgmt.py
```diff
@@ -582,7 +585,9 @@ def OpenNewRecordFile(dirIndex):
...
+                DebugPrintTraceback()
```
brianhlin, line 590:
```suggestion
                DebugPrintTraceback(0)
```

Can't we just use this instead of the DebugPrint above? From https://github.com/opensciencegrid/gratia-probe/blob/2.x/common/gratia/common/debug.py#L162-L166

brianhlin (commented):
Mostly LGTM. Nitpicks and a question.

common/gratia/common/sandbox_mgmt.py
djw8605: Yup, that's fine.  I didn't read much into the DebugTraceback.  Though, the debug print does give some "context" to the exception.

brianhlin:
```suggestion
```

Good point on the additional context. Let's just get rid of it since it's superfluous.

PR #170, opened 2023-03-15
Log and bail when there are dir creation issues (SOFTWARE-5531)

brianhlin:
Includes #169 . I've opted to just go with the exception route -- there aren't that many call sites:

1. https://github.com/opensciencegrid/gratia-probe/blob/2.x/common/gratia/common/send.py#L305-L305
2. https://github.com/opensciencegrid/gratia-probe/blob/2.x/common/gratia/common/send.py#L91-L91

But `Send` has quite a bit of call sites so I think I prefer just bailing loudly with an exception in this case because if the working dir isn't writeable then there's going to be a lot of trouble.

### DIFF
--- a/common/gratia/common/sandbox_mgmt.py
+++ b/common/gratia/common/sandbox_mgmt.py
@@ -569,8 +569,8 @@ def OpenNewRecordFile(dirIndex):
             if not os.path.exists(working_dir):
                 try:
                     Mkdir(working_dir)
-                except Exception as e:
-                    DebugPrint(0, 'Warning: Exception caught while creating directory: ' + working_dir + ':' + e)
+                except Exception as exc:
+                    DebugPrint(0, 'Warning: Exception caught while creating directory: ' + working_dir + ':' + exc)
                     continue
             if not os.path.exists(working_dir):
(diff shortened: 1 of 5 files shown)


common/gratia/common/sandbox_mgmt.py
```diff
@@ -569,21 +569,29 @@ def OpenNewRecordFile(dirIndex):
             if not os.path.exists(working_dir):
                 try:
                     Mkdir(working_dir)
-                except:
-                    continue
+                except Exception as exc:
+                    msg = 'Warning: Exception caught while creating directory: ' + working_dir
```
matyasselmeci:
```suggestion
                    msg = 'Fatal: Exception caught while creating directory: ' + working_dir
```
if you're bailing, it's not a warning.

brianhlin: Good point

brianhlin:
Since we're giving ourselves a bit more time on this release, I want to just try/except the write. marking `do-not-merge`

brianhlin:
@matyasselmeci it felt good to get rid of that. I'm planning on adding a linting GHA just to doubly make sure we're not shooting ourselves in the foot. I can include that in this PR or a separate one.

matyasselmeci (approved):
yay!

PR #172, opened 2023-03-23
Drop privs if running as root (SOFTWARE-5531)

brianhlin:
I'm not quite sure if we need to care about the case of someone running HTCondor/HTCondor-CE as something other than the `condor` user

### DIFF
--- a/condor-ap/condor_meter
+++ b/condor-ap/condor_meter
@@ -212,8 +212,7 @@ def main():
         try:
             become_condor()
         except (KeyError, PermissionError) as exc:
-            print(exc, file=sys.stderr)
-            sys.exit(1)
+            sys.exit(exc)
 
     try:
         opts, dirs = parse_opts()
--- a/condor-ap/condor_meter
+++ b/condor-ap/condor_meter
@@ -201,7 +201,10 @@ def become_condor(condor_service: str = ""):
             os.environ.setdefault("CONDOR_CONFIG", "/etc/condor-ce/condor_config")
             htcondor.reload_config()
 
-        condor_uid, condor_gid = htcondor.param['CONDOR_IDS'].split(".")
+        # Admins can specify the UID/GID to run the condor daemons as through the config "CONDOR_IDS = <UID>.<GID>",
+        # e.g. CONDOR_IDS = 1.2
+        condor_ids = htcondor.param['CONDOR_IDS']
+        condor_uid, condor_gid = [int(x) for x in condor_ids.split(".")]
     except KeyError:
         # When custom HTCondor IDs are not specified, they are using the 'condor' user
         try:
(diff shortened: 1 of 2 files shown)


condor-ap/condor_meter
```diff
@@ -189,8 +190,31 @@ def get_num_procs(job_ad):
...
+
+
 condor_history_re = re.compile("^history.(\d+)\.(\d+)")
 def main():
+    if os.getuid() == 0:
+        try:
+            become_condor()
+        except (KeyError, PermissionError) as exc:
+            print(exc, file=sys.stderr)
+            sys.exit(1)
```
matyasselmeci:
```suggestion
            sys.exit(exc)
```
bet you didn't know that was a thing, huh?

condor-ap/condor_meter
```diff
@@ -189,8 +190,31 @@ def get_num_procs(job_ad):
 
     return procs
 
+
+def become_condor():
```
matyasselmeci: Maybe try [CONDOR_IDS](https://htcondor.readthedocs.io/en/latest/admin-manual/configuration-macros.html#CONDOR_IDS) before the hardcoded 'condor' user.

matyasselmeci (commented):
See my comment about CONDOR_IDS

condor-ap/condor_meter
brianhlin: Oh, I use it all the time but didn't realize that it prints to stderr!

condor-ap/condor_meter
brianhlin: Resolved.

brianhlin:
I do want to update the tests in light of [a191587](https://github.com/opensciencegrid/gratia-probe/pull/172/commits/a1915872467da00419558a7771e9089c56e8e3a7) so let's hold off on merging this for now

condor-ap/condor_meter
```diff
@@ -189,8 +191,41 @@ def get_num_procs(job_ad):
...
+def become_condor(condor_service: str = ""):
+    """condor_meter is intended to run as a SchedD cron, i.e. as the 'condor' user.
+    condor_service: if 'htcondor-ce' determine the condor user for HTCondor-CE, otherwise use the default condor config
+    """
+    try:
+        if condor_service == 'htcondor-ce':
+            os.environ.setdefault("CONDOR_CONFIG", "/etc/condor-ce/condor_config")
+            htcondor.reload_config()
+
+        condor_uid, condor_gid = htcondor.param['CONDOR_IDS'].split(".")
```
matyasselmeci:
```suggestion
        try:
            condor_uid, condor_gid = htcondor.param['CONDOR_IDS'].split(".", 1)
            condor_uid, condor_gid = int(condor_uid), int(condor_gid)
        except (IndexError, ValueError):
            raise utils.InternalError("ERROR: malformed CONDOR_IDS setting")
```
at the very least you'll need the int conversion.  You might also want to handle CONDOR_IDS being defined but empty.

brianhlin: Oops, I didn't see that you made a suggestion in the email. I did the bare split on `.`:  I'd rather throw the unpack `ValueError` if there are more than one `.` separator

matyasselmeci (approved):
Nice!

PR #173, opened 2023-03-27
Refactor condor ID detection, reorganize gratia libs, bump version (SOFTWARE-5531)

brianhlin:
In SOFTWARE-5531, we decided to chown dirs properly in %post and I didn't want to reimplement the logic for how we determine the condor IDs. I factored out ID detection into a separate function and reworked the tests. To actually be able to call that function from outside of the script, I moved everything into a lib under `gratia-probe-common` and had the scripts just wrap around to call it.

You'll probably want to look at the commits outside of the rename for cleaner diffs.


condor-ap/condor_meter
```diff
@@ -0,0 +1,23 @@
...
+# htcondor-ce and condor-probe cases
+probe_name = os.path.basename(os.path.dirname(os.path.abspath(__file__)))
+
+if __name__ == "__main__":
+    try:
+        condor.main(probe_name)
+    except SystemExit:
+        raise
+    except KeyboardInterrupt:
+        raise
```
matyasselmeci: This can be deleted: as of Python 2.6 IIRC, SystemExit and KeyboardInterrupt derive from BaseException, not Exception.

rpm/gratia-probe.spec
```diff
@@ -449,6 +456,12 @@ The HTCondor-CE probe for the Gratia OSG accounting system
 %post htcondor-ce
 %customize_probeconfig -d htcondor-ce
 
+# Fix directory ownership if users have run condor_meter as root (SOFTWARE-5531)
+condor_ids=$(python3 -c 'import gratia.common.condor as condor; \
+             print(":".join([str(x) for x in core.get_condor_ids("htcondor-ce")]))')
+for dirname in $(awk -F '=' '/Folder=/ {print $2}' /etc/gratia/htcondor-ce/ProbeConfig | tr -d \"); do
```
matyasselmeci: This is terrifying. If someone messes up their ProbeConfig then the next time Gratia gets upgraded it will recursively chown whatever they put in there, which could be very destructive.

You should chown the default directories only (and declare ownership of them in the %files list if you haven't already); if the admin decides to move them elsewhere, that's their problem.

matyasselmeci (changes requested):
See my comment about chowning dirs in %post.

rpm/gratia-probe.spec
brianhlin: What's the worst that could happen 😈 . Fixed.

matyasselmeci (approved):

PR #174, opened 2023-04-27
Fix NameError (Software-5558)

jeff-takaki:
Fixes NameError where core was not defined. Condor is the alias of import gratia.common.condor so core was replaced with condor in order to refer to that module properly.


brianhlin (approved):
LGTM!

PR #175, opened 2023-06-14
Fix GHA deprecation warning (Software-5354)

jeff-takaki:
Fix set-output GHA deprecation warning in build-sw-container.yml


brianhlin (approved):
LGTM

PR #176, opened 2023-07-31
Add Unit Test for CompressOutbox to Tempfile Creation Refactoring (SOFTWARE-5540)

jeff-takaki:
@brianhlin I have been getting `AssertionError: False is not true` and I think it has to do with how I am using the mock patch. I tried a few different ways of incorporating the mock with not much success. It's possible I am provisioning the test environment incorrectly??

### DIFF
--- a/test/test_sandbox_mgmt.py
+++ b/test/test_sandbox_mgmt.py
@@ -1,6 +1,8 @@
 #!/bin/env python
 
 import os
+import shutil
+import tempfile
 import unittest
 from unittest.mock import patch, PropertyMock
 
@@ -32,3 +34,34 @@ class SandboxMgmtTests(unittest.TestCase):
             except (FileNotFoundError, NameError):
                 # don't need to clean up what's not there
                 pass
+
+class CompressOutboxTests(unittest.TestCase):
+    def setUp(self):
+        #provision test environment
+        self.probe_dir = tempfile.mkdtemp()
+        self.outbox = os.path.join(self.probe_dir, 'outbox')
+        os.makedirs(self.outbox, exist_ok=True)
+        self.outfiles = ['testfile1', 'testfile2']
+        # add content to the files
+        for fname in self.outfiles:
+            with open(os.path.join(self.outbox, fname), 'w') as f:
+                f.write('test content')
+
+    def tearDown(self):
+        #Remove probe_dir after test
+        shutil.rmtree(self.probe_dir)
+
+    @patch('gratia.common.config.ConfigProxy.get_GratiaExtension', create=True, return_value ='test-extension')
+    @patch('gratia.common.config.ConfigProxy.getFilenameFragment', create=True, return_value ='test-filename')
+    def test_compress_outbox(self, mock_gratia_ext, mock_file_frag):
+        """CompressOutbox compresses the files in the outbox directory
+            and stores the resulting tarball in probe_dir/staged.
+        """
+        #Parameters for function
+        probe_dir = self.probe_dir
+        outbox = self.outbox
+        outfiles = self.outfiles
+
+        #Assert that CompressOutbox returns True
+        result = sandbox_mgmt.CompressOutbox(probe_dir, outbox, outfiles)
+        self.assertTrue(result)
--- a/test/test_sandbox_mgmt.py
+++ b/test/test_sandbox_mgmt.py
@@ -55,7 +55,7 @@ class CompressOutboxTests(unittest.TestCase):
     @patch('gratia.common.config.ConfigProxy.getFilenameFragment', create=True, return_value ='test-filename')
     def test_compress_outbox(self, mock_gratia_ext, mock_file_frag):
         """CompressOutbox compresses the files in the outbox directory
-            and stores the resulting tarball in probe_dir/staged.
+        and stores the resulting tarball in probe_dir/staged.
         """
         #Parameters for function
         probe_dir = self.probe_dir
--- a/test/test_sandbox_mgmt.py
+++ b/test/test_sandbox_mgmt.py
@@ -12,22 +12,23 @@ class SandboxMgmtTests(unittest.TestCase):
     def test_GenerateFilename(self, mock_config):
         """GenerateFilename creates a temporary file and returns the path to the file
         """
-        prefix = 'test-prefix.'
+        prefix = 'test-prefix'
         temp_dir = '/tmp'
 
         try:
-            filename = sandbox_mgmt.GenerateFilename(prefix, temp_dir)
-            self.assertTrue(os.path.exists(filename),
-                            f'Failed to create temporary file ({filename})')
-            self.assertEqual(temp_dir.rstrip('/'),
-                             os.path.dirname(filename),
-                             f'Temporary file {filename} placed in the wrong directory')
-            self.assertRegex(filename,
-                             rf'{temp_dir}/*{prefix}\d+\.{mock_config.return_value}__\w+',
-                             'Unexpected file name format')
+            with sandbox_mgmt.GenerateFilename(prefix, temp_dir) as filename:
+                self.assertTrue(os.path.exists(filename.name),
+                                f'Failed to create temporary file ({filename.name})')
+                self.assertEqual(temp_dir.rstrip('/'),
+                                 os.path.dirname(filename.name),
+                                 f'Temporary file {filename.name} placed in the wrong directory')
+                self.assertRegex(filename.name,
+                                 rf'{temp_dir}/*{prefix}\.\d+\.{mock_config.return_value}__\w+',
+                                 'Unexpected file name format')
         finally:
             try:
-                os.remove(filename)
-            except FileNotFoundError:
+                filename.close()
+                os.remove(filename.name)
+            except (FileNotFoundError, NameError):
                 # don't need to clean up what's not there
                 pass
--- a/test/test_sandbox_mgmt.py
+++ b/test/test_sandbox_mgmt.py
@@ -39,7 +39,7 @@ class SandboxMgmtTests(unittest.TestCase):
 
 class CompressOutboxTests(unittest.TestCase):
     def setUp(self):
-        #provision test environment
+        # provision test environment
         self.probe_dir = tempfile.mkdtemp()
         self.outbox = os.path.join(self.probe_dir, 'outbox')
         os.makedirs(self.outbox, exist_ok=True)
--- a/test/test_sandbox_mgmt.py
+++ b/test/test_sandbox_mgmt.py
@@ -40,81 +40,98 @@ class SandboxMgmtTests(unittest.TestCase):
 class CompressOutboxTests(unittest.TestCase):
     def setUp(self):
         # provision test environment
+        gratia_ex = patch('gratia.common.config.ConfigProxy.get_GratiaExtension',
+                          create=True, return_value='test-extension')
+        file_frag = patch('gratia.common.config.ConfigProxy.getFilenameFragment',
+                          create=True, return_value='test-filename')
+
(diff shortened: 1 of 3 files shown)


brianhlin:
@jtakaki-matc I don't see anything obvious that would indicate why this failed so I would

1. Comment out the teardown function
2. Run this test locally and inspect the filesystem to see if the expected tarball gets created. This may be a good spot to add some changes to the actual `CompressOutbox` function to add debugging `print()` statements

jeff-takaki:
@brianhlin Test seems to be working, checked locally running from 2.x branch as well.

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,67 @@
...
+    def tearDown(self):
+        #Remove probe_dir after test
+        shutil.rmtree(self.probe_dir)
+
+    @patch('gratia.common.config.ConfigProxy.get_GratiaExtension', create=True, return_value ='test-extension')
+    @patch('gratia.common.config.ConfigProxy.getFilenameFragment', create=True, return_value ='test-filename')
+    def test_compress_outbox(self, mock_gratia_ext, mock_file_frag):
+        """CompressOutbox compresses the files in the outbox directory
+            and stores the resulting tarball in probe_dir/staged.
+        """
```
brianhlin:
```suggestion
        """CompressOutbox compresses the files in the outbox directory
        and stores the resulting tarball in probe_dir/staged.
        """
```

This multiline string is printed as is (try it in your python interpreter) so your extra 4 spaces will be included and look something like:

```
$ python
Python 3.11.4 (main, Jun  7 2023, 00:00:00) [GCC 12.3.1 20230508 (Red Hat 12.3.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print("""CompressOutbox compresses the files in the outbox directory
...     and stores the resulting tarball in probe_dir/staged.
... """)
CompressOutbox compresses the files in the outbox directory
    and stores the resulting tarball in probe_dir/staged.

```

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,67 @@
...
+        #Parameters for function
+        probe_dir = self.probe_dir
+        outbox = self.outbox
+        outfiles = self.outfiles
+
+        #Assert that CompressOutbox returns True
+        result = sandbox_mgmt.CompressOutbox(probe_dir, outbox, outfiles)
+        self.assertTrue(result)
```
brianhlin, line 88: A few weeks back we had a discussion about the different things that this function does: returns a boolean but also a tarball gets created in an expected location made up of the specified outfiles. It may be useful to test for things like:

1. The tarball is created in the expected location with expected pattern for the filename (similar to assertions in https://github.com/opensciencegrid/gratia-probe/pull/176/files#diff-a41929032101b1500faa4564484c19a2829fb12c86cd3b95182aa9a66e5071daR22-R29)
2. The unpacked tarball contents are the same as what's given in the outfiles

I'd probably organize these as separate tests and you could even refactor the call to the `CompressOutbox` function into the `setUp()` test!

brianhlin (changes requested):
This is looking great! One minor whitespace change and a request for additional tests.

EDIT: Could you also update the title of this PR? I don't think it summarizes ALL of the commits in the PR

jeff-takaki:
@brianhlin Added two more tests!

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+            try:
+                filename.close()
+                os.remove(filename.name)
+            except (FileNotFoundError, NameError):
+                # don't need to clean up what's not there
+                pass
+
+class CompressOutboxTests(unittest.TestCase):
+    def setUp(self):
+        #provision test environment
```
brianhlin:
```suggestion
        # provision test environment
```

https://www.flake8rules.com/rules/E265.html. Make sure this gets applied to your other comments, too!

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+
+        #Finds tarball that matches GenerateFilename function output
+        os.chdir(f'{probe_dir}/staged/store')
+        for tarball in glob.glob("tz.*.test-extension__*"): 
+            return(tarball)
+        
+        self.assertTrue(os.path.exists(path_to_tarball/{tarball}))
+
+    @patch('gratia.common.config.ConfigProxy.getFilenameFragment', create=True, return_value ='test-filename')
+    @patch('gratia.common.config.ConfigProxy.get_GratiaExtension', create=True, return_value ='test-extension')
```
brianhlin: Since you're applying these patches to every single test, you can refactor them out into `setUp()`. See https://github.com/opensciencegrid/gratia-probe/blob/2.x/test/test_condor.py#L126-L137 for an example

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        #Parameters for function
+        probe_dir = self.probe_dir
+        outbox = self.outbox
+        outfiles = self.outfiles
+
+        sandbox_mgmt.CompressOutbox(probe_dir, outbox, outfiles)
+
+        #Finds tarball that matches GenerateFilename() output
+        os.chdir(f'{probe_dir}/staged/store')
```
brianhlin: It's a little silly to define new variables for so few call sites. Just reference the `self.*` variables directly. That's what they're there for!

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+        probe_dir = self.probe_dir
+        outbox = self.outbox
+        outfiles = self.outfiles
+
+        sandbox_mgmt.CompressOutbox(probe_dir, outbox, outfiles)
+
+        #Finds tarball that matches GenerateFilename function output
+        os.chdir(f'{probe_dir}/staged/store')
+        for tarball in glob.glob("tz.*.test-extension__*"): 
+            return(tarball)
```
brianhlin: Why return here? This will stop the test before your assertion below

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+        
+        self.assertTrue(os.path.exists(path_to_tarball/{tarball}))
```
brianhlin: This is the first reference to `path_to_tarball` that I can see so this will throw an exception. Also the `{}` formatting looks off. I'd suggest getting pylint/flake8 set up in VSCode so that it's easier to catch these sorts of issues

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+        #Parameters for function
+        probe_dir = self.probe_dir
+        outbox = self.outbox
+        outfiles = self.outfiles
+
+        sandbox_mgmt.CompressOutbox(probe_dir, outbox, outfiles)
+
+        #Finds tarball that matches GenerateFilename function output
+        os.chdir(f'{probe_dir}/staged/store')
+        for tarball in glob.glob("tz.*.test-extension__*"):
```
brianhlin: This is a good catch that the glob may return multiple values. Since the function should NOT create multiple tarballs, you should add an assertion here that you only get one tarball back

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+        #Finds tarball that matches GenerateFilename() output
+        os.chdir(f'{probe_dir}/staged/store')
+        for tarball in glob.glob("tz.*.test-extension__*"):
+            return(tarball)
+
+        #Gets names of files within tarball
+        file_obj= tarfile.open(tarball,"r")
+        namelist=file_obj.getnames()
+        for names in namelist:
+            return(names)
```
brianhlin: Same comment about a strange return in here

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+        file_obj.close()
```
brianhlin: I think you can avoid the `close()` here if you use a `with` wrapper

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # Sort both lists to ensure order-independent comparison
+        names.sort()
+        outfiles.sort()
+
+        self.assertListEqual(names, outfiles)
```
brianhlin: This is a good assertion! Another one would be to also check that the contents are expected

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,120 @@
...
+    def setUp(self):
+        #provision test environment
+        self.probe_dir = tempfile.mkdtemp()
+        self.outbox = os.path.join(self.probe_dir, 'outbox')
+        os.makedirs(self.outbox, exist_ok=True)
+        self.outfiles = ['testfile1', 'testfile2']
+        # add content to the files
+        for fname in self.outfiles:
+            with open(os.path.join(self.outbox, fname), 'w') as f:
+                f.write('test content')
```
brianhlin: It would be good to make the contents specific per file (perhaps insert the filename at the front?)

brianhlin (changes requested):
A couple of improvements to be made but this is looking great!

jeff-takaki:
@brianhlin Made a few changes per your suggestions.

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,137 @@
...
+        # Finds exactly one tarball that matches GenerateFilename function output
+        tarball = glob.glob("tz.*.test-extension__*", root_dir=path_to_tarball)
+        tarball_location = os.path.join(f'{path_to_tarball}', tarball[0])
+        # Counts files in the directory that the tarball should be in
+        tarball_count = len((tarball))
+
+        self.assertTrue(os.path.exists(tarball_location),
+                        'Tarball not created in correct location')
+        self.assertEqual(tarball_count, 1,
+                         'Tarball created in directory != 1')
```
brianhlin:
```suggestion
                         f'Expected 1 tarball, found {tarball_count}')
```

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Assert that tarball is created in the correct location
+        """
+        sandbox_mgmt.CompressOutbox(self.probe_dir, self.outbox, self.outfiles)
+
+        path_to_tarball = f'{self.probe_dir}/staged/store'
+
+        # Finds exactly one tarball that matches GenerateFilename function output
+        tarball = glob.glob("tz.*.test-extension__*", root_dir=path_to_tarball)
+        tarball_location = os.path.join(f'{path_to_tarball}', tarball[0])
```
brianhlin: What happens when you find 0 tarballs?

brianhlin (approved):
Looks solid! A suggestion to improve the message upon one test failure and a slight logic issue. Pre-approving

test/test_sandbox_mgmt.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Assert that unpacked tarball contains files from outfiles
+        """
+        sandbox_mgmt.CompressOutbox(self.probe_dir, self.outbox, self.outfiles)
+        path_to_tarball = f'{self.probe_dir}/staged/store'
+
+        # Finds tarball that matches GenerateFilename() output
+        tarball = glob.glob("tz.*.test-extension__*", root_dir=path_to_tarball)
+        # Where tarball exists
+        tarball_location = os.path.join(f'{path_to_tarball}', tarball[0])
```
brianhlin: I just noticed: this will also blow up if there are 0 tarballs

jeff-takaki:
@brianhlin Refactored!

PR #183, opened 2024-08-06
Add script to convert kapel output records to Gratia records 

mwestphall:
- Add a new docker image that can be used as the second stage in the [kapel](https://github.com/rptaylor/kapel) processing pipeline


.github/workflows/build-k8s-container.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+    - name: make tags
+      id: mktags
+      env:
+        DTAG=${{ steps.mkdatetag.outputs.dtag }}
+      run: |
+        reponame="osg-htc/gratia-probe-k8s"
+        tags=()
+        for registry in docker.io hub.opensciencegrid.org; do
+            tags+=("$registry/$reponame:release")
```
brianhlin: Since we pull things out of OSG 23, we should probably align tags, e.g. `23-release`

kubernetes/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Useful for handling configuration
+environs
+# Useful for adding messages to outgoing queue
+dirq
```
brianhlin: Can / should we version lock these?

kubernetes/kubernetes_meter.py
```diff
@@ -0,0 +1,176 @@
...
+    for name in queue:
+        records.append(name)
+        if len(records) >= batch_size:
+            yield records
+            records = []
+    # Yield the last entries as well
+    yield records
+
+def main(envFile: str):
+    print(f'Starting Gratia post-processor: {__file__} with envFile {envFile} at {datetime.now(tz=timezone.utc).isoformat()}')
```
brianhlin, line 145: Can we also dump the contents of the env file? It could be helpful for debugging.

brianhlin (approved):
Minor comments but otherwise LGTM

PR #186, opened 2024-08-29
Kuantifier Release Candidate Bugfixes

mwestphall:
- Handle printing the env from a `None` envfile argument, which is the default
- Update parsing of APEL records to handle cases where `:` is in the value (eg. `SubmitHost: http://localhost:8000`)
- Round up processor count to the nearest integer rather than down (so 7.9 becomes 8 instead of 7)

### DIFF
--- a/kubernetes/kubernetes_meter.py
+++ b/kubernetes/kubernetes_meter.py
@@ -44,9 +44,10 @@ class ApelRecordConverter():
         self.apel_dict = {}
         lines = apel_record.split('\n')
         for line in lines:
-            kv_pair = [v.strip() for v in line.split(':')]
-            if len(kv_pair) == 2:
-                self.apel_dict[kv_pair[0]] = kv_pair[1]
+            # Handle case where apel values contain colons (eg. 'SubmitHost: http://localhost:8000')
+            kv_pair_match = re.match(r'([A-Za-z0-9_]+)\s*:\s*([^\s].*[^\s])\s*',line)
+            if kv_pair_match:
+                self.apel_dict[kv_pair_match[1]] = kv_pair_match[2]
 
     def getint(self, key):
         return int(float(self.apel_dict.get(key, 0)))
--- a/kubernetes/kubernetes_meter.py
+++ b/kubernetes/kubernetes_meter.py
@@ -4,6 +4,7 @@ import datetime
 from gratia_k8s_config import GratiaK8sConfig
 from dirq.QueueSimple import QueueSimple
 import os
+from math import ceil
 from datetime import datetime, timezone 
 import re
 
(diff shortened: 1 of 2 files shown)


kubernetes/kubernetes_meter.py
```diff
@@ -44,13 +44,17 @@ def _parse_apel_str(self, apel_record: str):
...
+            kv_pair_match = re.match(r'([A-Za-z0-9_]+)\s*:\s*([^\s].*[^\s])\s*',line)
+            if kv_pair_match:
+                self.apel_dict[kv_pair_match[1]] = kv_pair_match[2]
 
     def getint(self, key):
         return int(float(self.apel_dict.get(key, 0)))
 
+    def getint_roundup(self, key):
+        return int(float(self.apel_dict.get(key, 0)) + 0.5)
+
```
brianhlin: Does this actually work? 

```
blin@blin-work:~$ python
Python 3.12.3 (main, Apr 17 2024, 00:00:00) [GCC 13.2.1 20240316 (Red Hat 13.2.1-7)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> int(float(3.9))
3
```

Are you looking for `math.ceiling()` instead?

kubernetes/kubernetes_meter.py
```diff
@@ -44,13 +44,17 @@ def _parse_apel_str(self, apel_record: str):
         self.apel_dict = {}
         lines = apel_record.split('\n')
         for line in lines:
-            kv_pair = [v.strip() for v in line.split(':')]
-            if len(kv_pair) == 2:
-                self.apel_dict[kv_pair[0]] = kv_pair[1]
+            # Handle case where apel values contain colons (eg. 'SubmitHost: http://localhost:8000')
+            kv_pair_match = re.match(r'([A-Za-z0-9_]+)\s*:\s*([^\s].*[^\s])\s*',line)
```
brianhlin: How about just limiting the number of splits?

```
>>> 'SubmitHost: http://localhost:8000'.split(':', maxsplit=1)
['SubmitHost', ' http://localhost:8000']
```

brianhlin (approved):
Some ideas -- preapproving

kubernetes/kubernetes_meter.py
mwestphall: This rounds >= X.5 to X +1, <= X.5 to X, eg:
```
>>> int(float("3.9") + .5)
4
>>> int(float("3.4") + .5)
3
```

. math.ceiling() is probably more appropriate though

kubernetes/kubernetes_meter.py
mwestphall: Didn't know about the maxsplit kwarg, that's very useful

kubernetes/kubernetes_meter.py
brianhlin: Ah duh, that makes sense. `math.ceiling` doesn't do any sort of rounding so maybe what you had was right?

I'm not actually sure I understand what the ultimate goal is here.

mwestphall: CPU requests are configured as 0.9 and 7.9 for 1 and 8 core jobs respectively in Armen's cluster to allow other small pods to run on the same node, which were getting truncated to 0 or 7 by this code. Either math.ceil or the +.5 trick will round them to 1 and 8 as desired in this case. I think math.ceil is slightly preferable since it will round a processor count of <0.5 to 1 instead of 0, but that is probably a rare corner case.

brianhlin: Huh and why do we need to turn this into an int?

mwestphall: That is a good question. I looked at several sample gratia records while implementing this and noticed that all fields were integers, but never confirmed that that was actually a requirement

PR #189, opened 2024-12-04
SOFTWARE-6037: Update 50-gratia-gwms.conf with SCHEDD_CRON_LOG_NON_ZERO_EXIT

mwestphall:


brianhlin (approved):
Could you also reference `SOFTWARE-6037` in the comment? Doesn't hurt to have more pointers to reasoning, related work, etc

LGTM

PR #190, opened 2024-12-10
Bump gratia-probe.spec release number

mwestphall:


rpm/gratia-probe.spec
```diff
@@ -607,6 +607,9 @@ This is a transitional dummy package for gratia-probe-slurm; it may safely be re
 %files slurm
 
 %changelog
+* Tue Dec 10 2024 Matt Westphall <westphall@wisc.edu.> -  2.8.5-1
+- Log gratia probe failures to aid debugging (SOFTWARE-6037)
+
```
brianhlin: There's also the fix from https://opensciencegrid.atlassian.net/browse/SOFTWARE-5558

brianhlin (approved):
One missing changelog entry but otherwise LGTM. Pre-approving

rpm/gratia-probe.spec
mwestphall:
```suggestion
  - Fix NameError in post-install script (SOFTWARE-5558)
```

rpm/gratia-probe.spec
```diff
@@ -607,6 +607,9 @@ This is a transitional dummy package for gratia-probe-slurm; it may safely be re
 %files slurm
 
 %changelog
+* Tue Dec 10 2024 Matt Westphall <westphall@wisc.edu.> -  2.8.5-1
+- Log gratia probe failures to aid debugging (SOFTWARE-6037)
+  - Fix NameError in post-install script (SOFTWARE-5558)
```
mwestphall:
```suggestion
- Fix NameError in post-install script (SOFTWARE-5558)
```

PR #193, opened 2025-10-13
SOFTWARE-6232: htcondor2 bindings

mwestphall:


osg-pilot-container/osgpilot_meter
```diff
@@ -20,8 +20,8 @@ import gratia.common.Gratia as Gratia
 #import gratia.common.file_utils as file_utils
 import gratia.common.config as config
 
-import classad
-import htcondor
+import classad2 as classad
+import htcondor2 as htcondor
```
brianhlin:
```suggestion
import classad
import htcondor
```

Let's not make any functional changes until we understand behavior

osg-pilot-container/osgpilot_meter
```diff
@@ -295,6 +295,8 @@ def slottype_shortname(s):
 
 def eval_expr(x):
     if isinstance(x, classad.ExprTree):
+        # TODO: Per docs, eval() with no positional arg may yield unexpected results.
+        # We must confirm continued functionality.
```
brianhlin:
```suggestion
        # TODO: Per htcondor2 docs, eval() with no positional arg may yield unexpected results.
        # We must confirm continued functionality before updating
```

brianhlin (commented):
I'd prefer leaving the `osg-pilot-container` probe alone until we have a plan. Also, we should bump the condor requirement on `gratia-probe` condor subpackages so that they pull in a version with `htcondor2` available. I think that's 24.0.0 at a minimum but it's not clear from the upstream release notes when exactly they declared it ready.

brianhlin:
It looks like we need to import the htcondor2/classad2 python libraries into the tests

brianhlin:
@mwestphall : Jason had a good idea in that we could `try` importing the new libraries and fallback to the old ones if that doesn't work. That way we can 1) get rid of the version requirement and 2) ship this in 24-main

mwestphall:
@brianhlin good idea. I've updated the package proper to use try-importing, I don't think it will be necessary for the tests since they are now just pulling in the htcondor2 python packages.

brianhlin (approved):
LGTM

PR #194, opened 2026-01-26
Handle undefined resource names

dsschult:
If the resource name classad exists but is set to the value `classad.Value.Undefined` (we appear to have this when flocking to CHTC directly), allow processing to continue.  

Example of classads:

```
LastRemoteHost = "slot1_30@e2588.chtc.wisc.edu"
LastRemotePool = "cm.chtc.wisc.edu"
MATCH_EXP_JOBGLIDEIN_ResourceName = "Local Job"
MachineAttrGLIDEIN_ResourceName0 = undefined
```

Before this PR, it would look at `MachineAttrGLIDEIN_ResourceName0` first and return the undefined value, then break later in the code.  Now it says it's a local job, which seems better?

I leave it up to OSG to decide if they want CHTC to be counted.

### DIFF
--- a/common/gratia/common/condor.py
+++ b/common/gratia/common/condor.py
@@ -565,7 +565,7 @@ def cream_match(match, desired):
 def get_classad_resource_name(classad):
     for attr in RESOURCE_NAME_ATTRS:
         resource_name = classad.get(attr)
-        if resource_name:
+        if resource_name and resource_name not in (classadLib.Value.Undefined, classadLib.Value.Error):
             return resource_name
     return None
 


common/gratia/common/condor.py
```diff
@@ -565,7 +565,7 @@ def cream_match(match, desired):
 def get_classad_resource_name(classad):
     for attr in RESOURCE_NAME_ATTRS:
         resource_name = classad.get(attr)
```
mwestphall, line 567:
```suggestion
        resource_name = classad.eval(attr)
```

common/gratia/common/condor.py
```diff
@@ -565,7 +565,7 @@ def cream_match(match, desired):
...
-        if resource_name:
+        if resource_name and resource_name not in (classadLib.Value.Undefined, classadLib.Value.Error):
```
mwestphall, line 568:
```suggestion
        if resource_name and resource_name is not classadLib.Value.Undefined and resource_name is not classadLib.Value.Error:
```

mwestphall: Per the HTCondor dev team, you need to use identity comparisons to check against classadLib.Value values since they're just integers

common/gratia/common/condor.py
mwestphall: Eval is more robust than get here in case the classads are not literal values

mwestphall (commented):
Hi @dsschult , I've left a couple comments on this PR.

mwestphall (changes requested):
Requesting a couple changes

PR #199, opened 2026-05-12
Bump version to 2.9.2 and update release number

mwestphall:


rpm/gratia-probe.spec
```diff
@@ -607,6 +607,9 @@ This is a transitional dummy package for gratia-probe-slurm; it may safely be re
 %files slurm
 
 %changelog
+* Tue May 12 2026 Matt Westphall <westphall@wisc.edu.> -  2.9.2-1
+- Bump version number for code changes made in 2.9.1-3
```
brianhlin: Not part of this PR but can you add a more descriptive changelog than "Add missing try/catch block"? That communicates the exact changes made from the developer perspective but doesn't really tell sys admins what this change actually does. It should be along the lines of "Fix..." and then a summary of the bug

brianhlin (approved):
One comment about the changelog text but otherwise this LGTM. Pre-approving

